Issue #7, opened 2020-03-31
Add zone map to app


mjmaurer:
Preferably client side so they can used a cache version of the webapp and it will work if the backend is down

mjmaurer:
It should be added to [this page](crownheightsma.herokuapp.com/neighborhood-finder).

The code for that [page is here](https://github.com/crownheightsaid/mutual-aid-app/blob/master/src/webapp/pages/NeighborhoodFinder.js#L64)!

The geojson for the zone map is here: https://github.com/crownheightsaid/mapbox/blob/master/crownheights-quads.geojson

mjmaurer:
Bonus points if you can add the location to the map when it is entered!

piratefsh:
heck yeah i can take this!

mjmaurer:
lemme know if you need any pointers! I was thinking off to the side for desktop and below for mobile

PR #21, opened 2020-04-03
Add zone map to neiborhood finder

piratefsh:
Closes https://github.com/crownheightsaid/mutual-aid-app/issues/7

## Changes
Displays quadrant map and a marker of current address entered. Handles cases where address entered is outside of Crown Heights.

Sorry its kinda ugly right now, I'm just using whatever mapbox style is available. I chose mapbox only because historically it has a more flexible API to work with and is better documented.

### Default state:
Screen Shot 2020-04-04 at 12 51 25 PM

### Address (in Crown Heights) entered:
Screen Shot 2020-04-04 at 12 52 58 PM

### Address (outside of CH):
Zooms to fit bounds with address marker
Screen Shot 2020-04-04 at 12 53 14 PM

### Viewport < md
Screen Shot 2020-04-04 at 12 51 29 PM

## Summary
- add title
- return location data in `geo.js` endpoint
- add QuadrantMap component
- make page two layout and responsive (1 column when < md viewport)
## Follow up
- Add REACT_APP_MAPBOX_ACCESS_TOKEN to heroku prod environment. I have already added it to staging. I don't have prod heroku access so someone needs to do it here.

### DIFF
--- a/src/api/geo.js
+++ b/src/api/geo.js
@@ -30,14 +30,18 @@ exports.addressHandler = async (req, res, next) => {
     const geoResults = geoResult.data.results;
 
     const locResult = geoResults[0];
+    const {
+      geometry: { location }
+    } = locResult;
+
     const neighborhood = locResult.address_components.find(component =>
       component.types.includes("neighborhood")
     );
+
     const neighborhoodName = neighborhood ? neighborhood.long_name : "";
-    const [lt, long] = [
-      locResult.geometry.location.lat,
-      locResult.geometry.location.lng
-    ];
+
+    const [lt, long] = [location.lat, location.lng];
+
     const intersection = await geonamesClient.findNearestIntersection({
       lat: lt,
       lng: long
@@ -53,6 +57,7 @@ exports.addressHandler = async (req, res, next) => {
     return res.end(
       JSON.stringify({
         neighborhoodName,
+        location,
         intersection: {
           street_1: intersection.intersection.street1,
           street_2: intersection.intersection.street2
--- a/src/webapp/components/QuadrantMap.js
+++ b/src/webapp/components/QuadrantMap.js
@@ -1,0 +1,67 @@
+import React from "react";
+import ReactMapboxGl, { Feature, Layer, Source } from "react-mapbox-gl";
+import quadrantsGeoJSON from "../../assets/crownheights.json";
+
+const CROWN_HEIGHTS_CENTER_COORD = [-73.943018, 40.671254];
+
+const MapboxMap = ReactMapboxGl({
+  accessToken:
+    "pk.eyJ1IjoicGlyYXRlZnNoIiwiYSI6ImNrOGppNWJpcTBoM2wzbnIwb3kxcmNlcWYifQ.B8cTgg45CTfcEwSdOfHXrQ"
+});
+
+const QuadrantMap = ({ location }) => {
+  return (
+    <MapboxMap
+      style="mapbox://styles/mapbox/streets-v9"
+      center={CROWN_HEIGHTS_CENTER_COORD}
+      containerStyle={{
+        height: "300px",
+        width: "500px"
+      }}
+      zoom={[12]}
+    >
+      <Source
+        id="quadrants_src"
+        geoJsonSource={{
+          type: "geojson",
+          data: quadrantsGeoJSON
+        }}
+      />
+
+      <Layer
+        sourceId="quadrants_src"
+        type="fill"
+        paint={{
+          "fill-color": {
+            type: "identity",
+            property: "fill"
+          },
+          "fill-opacity": 0.4
+        }}
+      />
+      <Layer
+        sourceId="quadrants_src"
+        type="symbol"
+        layout={{
+          "text-field": {
+            type: "identity",
+            property: "id"
+          },
+          "text-size": 20
+        }}
+      />
+
+      {location && (
+        <Layer
+          type="symbol"
+          id="marker"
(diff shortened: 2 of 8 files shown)


src/webapp/pages/NeighborhoodFinder.js
```diff
@@ -10,6 +10,15 @@ import MailOutlineIcon from "@material-ui/icons/MailOutline";
 import InputAdornment from "@material-ui/core/InputAdornment";
 import Button from "@material-ui/core/Button";
 import Divider from "@material-ui/core/Divider";
+import ReactMapboxGl, { GeoJSONLayer, Layer, Source } from "react-mapbox-gl";
+import quadrantsGeoJSON from "../../assets/crownheights.json";
+
+const Map = ReactMapboxGl({
```
piratefsh, line 16: todo: fetch this from env

mjmaurer:
looking good !

src/webapp/components/QuadrantMap.js
```diff
@@ -0,0 +1,66 @@
+import React from "react";
+import ReactMapboxGl, { Feature, Layer, Source } from "react-mapbox-gl";
+import quadrantsGeoJSON from "../../assets/crownheights.json";
+
+const CROWN_HEIGHTS_CENTER_COORD = [-73.943018, 40.671254];
+
+const MapboxMap = ReactMapboxGl({
```
piratefsh, line 7: CRA fetches this from the env during build time: https://create-react-app.dev/docs/adding-custom-environment-variables/

src/api/geo.js
```diff
@@ -53,6 +57,7 @@ exports.addressHandler = async (req, res, next) => {
     return res.end(
       JSON.stringify({
         neighborhoodName,
+        location,
```
piratefsh, line 60: return latlng data in API

piratefsh: i did not see tests for these, please let me know if i should update them.

mjmaurer:
Thanks this looks great ! 

I think my only comment is on the map placement. Could we place it off to the side of the form?

Kinda like this: https://drive.google.com/file/d/1_shaGtUKjCMX4-CWyb63uF8Aivg-EIpp/view?usp=sharing

Style should look kinda like:
```
  root: {
    paddingTop: theme.spacing(3),
    paddingLeft: theme.spacing(3),
    paddingRight: theme.spacing(3),
    display: "flex";
    flex-wrap: "wrap";
    width: "100%"
  },
  formRoot: {
    flex: "0 0 65%";
  },
  mapRoot: {
    flex: 1;
  },
  field: {
    marginTop: theme.spacing(3),
    width: "85%"
  },
.....
```

src/webapp/helpers/mapbox-coordinates.js
```diff
@@ -0,0 +1,18 @@
+import { LngLat } from "mapbox-gl";
```
piratefsh, line 1: i should really test this, will TODO. for now they are simple enough that i am not too worried but yeah.

src/webapp/pages/NeighborhoodFinder.js
```diff
@@ -95,91 +101,102 @@ export default function NeighborhoodFinder() {
...
-            id="neighborhood"
-            label="Neighborhood"
-            value={data.neighborhoodName || "Unavailable"}
-            helperText="If both this and zone are unavailable, double check the map: https://bit.ly/2UrZPkA"
-            variant="outlined"
-            className={classes.field}
-          />
+      <Box className={classes.formRoot}>
+        <Box>
+          <Typography className={classes.text} variant="h3">
```
piratefsh, line 106: add heading for readability

src/webapp/pages/NeighborhoodFinder.js
```diff
@@ -95,91 +101,102 @@ export default function NeighborhoodFinder() {
...
+            The address will not be stored or logged :)
+          </Typography>
+        </Box>
+        <form onSubmit={handleSubmit} autoComplete="off">
           <TextField
-            disabled
-            id="zone"
-            label="Crown Heights Volunteer Zone"
-            value={data.quadrant || "Unavailable"}
+            id="address"
```
piratefsh, line 121: no changes here just moved into a wrapper Box

src/webapp/pages/NeighborhoodFinder.js
piratefsh: do you mind explaining why aim for minimal text? from a UI perspective, if users have a number of tabs open, it will be easier for them to pick out if they're on the right one if there's an obvious title. it's also helpful for those who are new to the tool.

src/webapp/components/QuadrantMap.js
piratefsh: sure can do

piratefsh:
@mjmaurer handled missing token state:
Screen Shot 2020-04-04 at 9 44 51 PM

src/webapp/pages/NeighborhoodFinder.js
mjmaurer: > if users have a number of tabs open, it will be easier for them to pick out if they're on the right one if there's an obvious title

I think this is more on the favicon, which is still the CRA one :/

Thought I think you're right. Can we just make it much smaller? h5? No need to update the screenshots imo

piratefsh: yeah agree re: the favicon. that would be a different task. i was thinking more if you're cycling through tabs. tbh if we're doing it the right way it should still be a h1 with a smaller font size but probably not worth twiddling that right now. pushed update!

PR #26, opened 2020-04-06
Add shortcut to draft a "Delivery Needed" post

alexquick:
## Overview

Right now, when intake is completed the intake volunteers manually piece together a message to post on slack to find delivery volunteers. This PR adds a new shortcut to slack that assists them when doing this. The workflow around this has three parts:
1. Activate the shortcut and then pick an open request. Currently this means that the `Request` is in `Delivery Needed` or `Dispatch Started` states AND has not already posted this message. The `Request`s are presented with the most-recently edited first, so this should make it easy for intake volunteers.
2. A draft message is suggested with the data of the `Request` that the intake volunteer has chosen. It can be edited, and at this point the volunteer can also pick a channel.
3. A preview is shown with the ability to edit again or submit the request
4. The message is sent to the chosen channel and a link is DM'd to the intake volunteer.
5. Afterwards, the status emoji of the message is kept synced up

## Deployment
Some config changes to the slack app need to be made for this to work in production/staging:
 - An interactive message Shortcut with `callbackId` = `select_delivery_needed_request`needs to be set up.
 - Scopes: `channels:history` `channels:write` `chat:write` `chat:write_public` `commands`

## Screenshots
<details><summary>Open</summary>
<p>

### Flow
draft_request
### Messages getting updated according to `Status` changes
update_status
### Message in context
image

</p>
</details>

## Sample Template
<details><summary>Open</summary>
<p>

```markdown
:red_circle: Hey <!channel> we have a new request to deliver groceries to our neighbor at Utica & Sterling in *NE Crown Heights*:
*Timeline*: not stated
*Neighborhood*: NE Crown Heights
*Need*: Groceries / Shopping
*Cross Streets*: Utica & Sterling
*Description*: flour, sugar, butter, milk, lysol, apples, oranges
*Code*: 8HII
*Want to volunteer to help our neighbor?* Comment on this thread and <@U010VK3Q821> will follow up with more details.
_Reminder: Please don’t volunteer for delivery if you have any COVID-19/cold/flu-like symptoms, or have come into contact with someone that’s tested positive._
```

</p>
</details>


src/slackapp/endpoints/interactivity.js
```diff
@@ -22,17 +24,19 @@ const atdViewOpenEntryId = "assign-to-delivery";
...
     type: "message_action",
-    callback_id: atdViewOpenEntryId
+    callbackId: atdViewOpenEntryId
   },
   atdViewOpen
 );
+
 slackInteractions.viewSubmission(
-  { callback_id: atdViewSubmissionCallbackId },
+  { callbackId: atdViewSubmissionCallbackId },
```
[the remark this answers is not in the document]
mjmaurer: really good catch!! sorry about that

src/slackapp/flows/createDeliveryRequest.js
```diff
@@ -0,0 +1,379 @@
...
+
+/**
+ * Adds the interaction handlers for this flow to the slack app.
+ * This flow handles the "Post Open Request" shortcut available for
+ * intake volunteers to draft and send a new post asking for delivery volunteers
+ * from a Request.
+ *
+ * Slack Setup:
+ * - An interactive message Shortcut with callbackId: select_delivery_needed_request configured in the slack app.
+ * - Scopes: channels:history channels:write chat:write chat:write_public commands
```
mjmaurer, line 28: I think we should keep these scopes in one file. Right now theres a partial list in https://github.com/crownheightsaid/mutual-aid-app/blob/master/SETUP.md that I need to update

src/slackapp/flows/createDeliveryRequest.js
```diff
@@ -0,0 +1,379 @@
...
+ *
+ * Look at `suggestedTemplate` if you want to change the text that is suggested in the message draft.
+ */
+module.exports.register = function register(slackInteractions) {
```
mjmaurer, line 32: no-action: gorgeous!

src/airtable.js
```diff
@@ -51,6 +52,27 @@ exports.findRequestByExternalId = async externalId => {
...
+  console.log(`Finding: ${formula}`);
+  try {
+    const requests = await base("Requests")
+      .select({
+        filterByFormula: formula
+      })
+      .all();
+    const notInSlack = r => {
+      const meta = JSON.parse(r.get("Meta"));
+      return meta.slack_ts === undefined;
```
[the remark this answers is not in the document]
mjmaurer: I think we need to leave the ability to post even if it already exists in slack. The state should be the only indicator. The reason being is that it's pretty common to cross post between channels

I think storing this in the meta field would be useful for the polling logic below, but:
1. might be better suited for a separate PR
2. Going with opt-in approach, would be better handled by the "Assign to Delivery" shortcut

I think this would remove the need for the slack channel + ts in the meta field. 
I know this is a big part of this PR. What are your thoughts?

src/slackapp/flows/createDeliveryRequest.js
```diff
@@ -0,0 +1,379 @@
...
+  });
+
+  // DM the user with a permalink
+  const permalink = await slackapi.chat.getPermalink({
+    channel: deliveryMessage.channel,
+    message_ts: deliveryMessage.ts
+  });
+  slackapi.chat.postMessage({
+    channel: payload.user.id,
+    text: `*${context.code}*: Submitted a new request for delivery to <#${deliveryMessage.channel}>\n${permalink.permalink}`,
```
mjmaurer, line 134: nit: submitted -> posted

src/slackapp/flows/createDeliveryRequest.js
```diff
@@ -0,0 +1,379 @@
...
+      emoji: true
+    },
+    blocks: [
+      {
+        block_id: "select_request",
+        type: "input",
+        element: {
+          action_id: "selected_request",
+          type: "radio_buttons",
+          initial_option: options[0],
```
mjmaurer, line 184: no-action: you're a king

src/slackapp/flows/createDeliveryRequest.js
```diff
@@ -0,0 +1,379 @@
...
+  }
+
+  const extraFields = [
+    ["Timeline", request.get("Time Sensitivity") || "When convenient"],
+    ["Neighborhood", quadrant],
+    ["Need", needs],
+    ["Cross Streets", streets],
+    ["Language", (request.get("Languages") || []).join("/")],
+    ["Description", request.get("Intake General Notes")],
+    ["Code", request.get("Code")]
```
mjmaurer, line 351: Know of any clever ways to make this easier to find with regex? My delivery shortcut regex is failing right now

mjmaurer:
Few followups:
1. No failover when a request field is missing in airtable
2. Filter to only requests for the submitting volunteer (see createDeliveryRequest..js in [this commit](https://github.com/crownheightsaid/mutual-aid-app/commit/9fa49c0eeaa969d3e81194fd486baaa2a05f0bcd)) 
3. (optional) - I think we should avoid posting the chat permalink. Intake vols tend to have a ton of active chats going on

src/airtable.js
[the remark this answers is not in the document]
alexquick: Shouldn't cross-posts be done by linking the post rather than duplicating it? Otherwise the threads get out of control. 
image

Regardless, yeah we can definitely make all of the updating contingent on actions in slack to start with. If it ends up that volunteers end up starting a message through this and then finishing up in airtable for some reason we can automate it.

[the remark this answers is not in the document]
alexquick: Though we would need the ability to mark a request as complete too to launch (because these can't be edited). Ideally I think there would be an action to do a raw edit as well for cases where the workflow actions we have don't match with reality?

src/slackapp/flows/createDeliveryRequest.js
alexquick: Looks like you fixed it?

src/slackapp/flows/createDeliveryRequest.js
alexquick: Removed permalink entirely anyway as you suggested.

src/slackapp/flows/createDeliveryRequest.js
alexquick: Thanks!

src/slackapp/flows/createDeliveryRequest.js
alexquick: Maybe `/code[^\w\d]+(?<code>[\w\d]{4})$/im`? That way you don't have to manually futz with the length of the capture?

src/airtable.js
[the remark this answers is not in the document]
mjmaurer: > Shouldn't cross-posts be done by linking the post rather than duplicating it?

Yes, but I'm not sure how many volunteers know how to do this. I guess worst case they can copy paste, but I'd prefer to get people used to using the shortcut

> If it ends up that volunteers end up starting a message through this and then finishing up in airtable for some reason we can automate it. Though we would need the ability to mark a request as complete too to launch (because these can't be edited).

The "completed" is actually "assigned" which can be triggered through the "Assign to Delivery shortcut" (not currently implemented)

> Ideally I think there would be an action to do a raw edit as well for cases where the workflow actions we have don't match with reality?

Definitely agree, but not a blocker for this

src/slackapp/flows/createDeliveryRequest.js
mjmaurer: Sounds good. Resolved

src/airtable.js
```diff
@@ -122,7 +147,7 @@ exports.findVolunteerByEmail = async email => {
 
 exports.findVolunteerById = async id => {
   try {
-    return [base("Volunteers").find(id), null];
+    return [await base("Volunteers").find(id), null];
```
alexquick, line 150: @mjmaurer I never met an `await` I couldn't miss either

alexquick:
I fixed the 3 issues you brought up in the followup, good looking out. The error handling was.... not ready for a live demo.

So I think the thing to decide on is if this is going to also change the slack message as the request moves through its lifecycle? I think it is cool, but if you think it's going to frustrate intake more than help we can definitely take it out. In that case I think we should add the ability to manually edit the message so that they can make the message using the tool and then take it from there even if they don't end up using slack shortcuts to assign to volunteer and so on.

alexquick:
Sure thing. I merged in the changes you had made on your branch so if you want to merge now can you back out https://github.com/crownheightsaid/mutual-aid-app/pull/26/commits/350d2b7ad2733e4efa7c09b4b134455b4b0d07d3, otherwise I can do it tonight.

PR #39, opened 2020-04-11
feat: move to separated airtable-change-detector package

mjmaurer:
Tested in staging.

No diff in funcitonality


src/workers/airtable-sync/worker.js
```diff
@@ -1,68 +1,54 @@
...
+          record.didChange(statusFieldName) ||
+          record.didChange(slackIdFieldName)
+        ) {
+          promises.push(updateMessageContent(record));
+        }
+        if (record.didChange(triggerBackfillFieldName)) {
+          promises.push(updateMessageContent(record));
+        }
+      });
+      return Promise.all(promises);
```
alexquick, line 49: Nice! The only downside I can think of is that we have to watch out for weird data races and are more likely to blow through Airtable's request limit (though we need to solve that more holistically anyhow).

mjmaurer: very true, I'll leave a note

PR #40, opened 2020-04-12
fix: automatic center bounds + eslint errors

mjmaurer:
@piratefsh  fyi. just doing some cleaning so we can share this with other MA groups more easily


piratefsh:
👍

PR #45, opened 2020-04-17
refactor: airtable tables into own file with fields schema as JSON

mjmaurer:
Moves table-specific airtable logic into separate table file.

Also introduces a schema as JSON that can be partially generated by running a script on the airtable API webpage.


src/lib/airtable/tables/__schemaGen.js
```diff
@@ -0,0 +1,30 @@
+// You can enter this in the console on:
```
piratefsh, line 1: im a little confused about that the inputs and outputs/sideeffects of this script is, can you elaborate?

mjmaurer: For sure. I'll add some more docs and work it around a bit. Very goofy hack

mjmaurer: done

mjmaurer:
I'm gonna go ahead and merge. Alex if you have any issues I'll be happy to make a change

I think having a top level fields object is a good balance between flexability, readability and protecting us from field naming bugs

src/lib/airtable/tables/__tableSchemaGen.js
```diff
@@ -0,0 +1,47 @@
...
+// Because this script isn't perfect and field names can be long,
+// you should go through and normalize field accessors manually.
+
+// Updates after the initial table creation should probably be
+// done manually. This could still be useful for grabbing long
+// select-field options as JSON.
+
+/* eslint-disable no-undef */
+tableNameToGenerate = "Requests";
+camelCase = s =>
```
alexquick, line 24: this is gross and I love it so much

Issue #49, opened 2020-04-20
make strings translatable


ericfg:
Going to have a hard time getting some groups I know to use this unless the core interface can be in something other than english (based on each user's choice)

mjmaurer:
agreed. i18n was originally included but structure changed very quickly and so was removed.

ericfg:
wow. For me this is a major block to using this tool. 
any thoughts about getting back to your initial idea?

mjmaurer:
https://www.i18next.com/ is what I originally setup.

It works well, but very configurable. I can try and find time this week to reintroduce it with the opinionated setup I had.

ericfg:
thanks. I have zero js experience or I'd try to help

mjmaurer:
#59 centralizes strings with i18next, which makes adding support for i18n easier. Next steps:
1. Dropdown in webapp for selecting language
2. Request locale for user in slackapp

piratefsh:
@mjmaurer can i close this issue? just wanna clean up our list of issues a bit.

mjmaurer:
While it's somewhat easy to finish implementing, this issue is still open. Need to have a language picker where we consume strings (webapp, slack, etc).

PR #59, opened 2020-04-25
feat: centralize strings with i18next. This can be used to support multiple languages in a future PR

mjmaurer:
See README.md and DEVELOPING.md

i18next strings are stored in JSON files per namespace per language. Namespaces correspond roughly to a package in the monorepo. where strings might have been shared between namespaces, they were instead kept separate so users of this repo can only focus on packages that they use.

i18next is initialized separately on the server side (filesystem to get JSON) and the webapp (HTTP to get the JSON). The initialization for both is centralized in i18nextInit.js.


mjmaurer:
#49

mjmaurer:
I'm going to merge this because it's blocking work for other aid groups!

PR #60, opened 2020-04-26
Add /delivery-needed API

piratefsh:
## Summary
- Add new endpoint for fetching all requests from Airtable that is marked `Delivery Needed` and has been posted to Slack.
- Add lru-cache to cache geo lookups.

I decided to use in-memory LRU caching as a quick solution rather than caching the lookups in Airtable. The reasons being:
a) I don't want to be making writes to Airtable as it might take awhile if there are many requests.
b) Will need to handle checking when the cross street data has been updated.
c) Cross streets data overlaps. Plus this geo data can be reused in other APIs if we store it this way.

I also decided to return data in GeoJSON format, storing any peripheral data in `properties.meta`. This will make it much easier to display on a map with clustering, plus since it's an open standard and compatible with many mapping tools, this API can be reused elsewhere.

Lookups are  set to cache for 7 days. 

### Manual testing
Start the server and visit http://localhost:5000/api/delivery-needed/requests.json.
On first visit, request will take longer ~300ms for about 5 requests, but will be much quicker ~100ms on subsequent visits.

Example payload:
```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          -73.94984500000001,
          40.6768827
        ]
      },
      "properties": {
        "title": "RRGLAK",
        "meta": {
          "Code": "RRGLAK",
          "Cross Street #1": "Dean St",
          "Cross Street #2": "Nostrand Ave",
          "Neighborhood Area (See Map)": "NW"
        }
      }
    },
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          -73.9308727,
          40.6714765
        ]
      },
      "properties": {
        "title": "HOJQJK",
        "meta": {
          "Code": "HOJQJK",
          "Cross Street #1": "Utica Ave",
          "Cross Street #2": "Sterling St",
          "Neighborhood Area (See Map)": "NE"
        }
      }
    }
}
```

### Performance
timing data on first hit:
Screen Shot 2020-04-26 at 11 49 20 AM

subsequent hits:
Screen Shot 2020-04-26 at 11 49 26 AM


src/api/delivery-needed/fetchCoordFromCrossStreets.js
```diff
@@ -0,0 +1,49 @@
+const { Client } = require("@googlemaps/google-maps-services-js");
+const { makeCache } = require("../../lib/cache/makeCache");
+
+let googleGeoClient;
+let cache;
+
+// cache lookups for 7 days (miliseconds)
+const CACHE_AGE = 7 * 24 * 60 * 60 * 1000;
```
piratefsh, line 8: this was rather arbitrary of a decision. LMK if you think it can be longer.

src/api/delivery-needed/index.js
```diff
@@ -0,0 +1,60 @@
...
+  } = fields;
+
+  const requestsWithCoordsPromises = requestObj.map(async r => {
+    let metaJSON = {};
+    let slackChannelId;
+    const location = await fetchCoordFromCrossStreets(
+      `
+      ${r.fields[crossStreetFirst]},
+      ${r.fields[crossStreetSecond]},
+      Brooklyn
```
piratefsh, line 29: I had to add Brooklyn here to restrict the results even more. Cross street data is too vague and sometimes returns results out of NY despite the locality setting.

src/api/delivery-needed/index.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+    return {
+      type: "Feature",
+      geometry: {
+        type: "Point",
+        coordinates: [location.lng, location.lat]
+      },
+      properties: {
+        title: r.fields[code],
+        meta: {
```
piratefsh, line 47: I am only returning the minimum amount of information right now. Will think about adding first names and grocery lists to humanize each request.

src/lib/airtable/tables/requests.js
```diff
@@ -1,6 +1,23 @@
 const { merge } = require("lodash");
 const { airbase } = require("~airtable/bases");
 
+const requestNotInSlack = r => {
```
piratefsh, line 4: refactored these out so I can reuse them

src/lib/airtable/tables/requests.js
```diff
@@ -49,6 +66,32 @@ exports.findRequestByExternalId = async externalId => {
   }
 };
 
+// Find open requests that area already in Slack
+exports.findDeliveryNeededRequests = async () => {
```
piratefsh, line 70: basically the same as `findOpenRequestsForSlack` but filters for results that are in Slack.

alexquick: Maybe this is indicating that there's a root `findOpenRequests` in our domain model that can be created and used in the two implementations?

```
let findOpenRequests = async () => { /*most of this without any slack filtering*/ }
let findDeliveryNeedeRequests = async () => { 
    let [results, err] = await findOpenRequests()
    return [results.filter(r => !requestNotInSlack(r)), err]
}

let findOpenRequestsForSlack = async () => { 
    let [results, err] = await findOpenRequests()
    return [results.filter(r => requestNotInSlack(r)), err]
}
```

src/api/delivery-needed/fetchCoordFromCrossStreets.js
```diff
@@ -0,0 +1,49 @@
...
+
+  let geoResult;
+  const cacheKey = makeCacheKey(address);
+
+  // look in cache
+  if (cache.peek(cacheKey)) {
+    geoResult = cache.get(cacheKey);
+    console.log("cache hit");
+  } else {
+    console.log("cache miss");
```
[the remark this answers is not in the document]
piratefsh: oh hah those were debug lines that i forgot to remove, but yeah on loggers. either one should work for our purposes.

src/api/delivery-needed/fetchCoordFromCrossStreets.js
piratefsh: gotcha, maybe i can set the cache to be 24 hours instead then. i did the math and if we have ~50 deliveries per day and even if requested it every 10mins, we'd still be way below the 100k limit.

src/api/delivery-needed/fetchCoordFromCrossStreets.js
```diff
@@ -0,0 +1,49 @@
...
+// cache lookups for 7 days (miliseconds)
+const CACHE_AGE = 7 * 24 * 60 * 60 * 1000;
+
+const makeCacheKey = address => {
+  return `geoCache-${address}`;
+};
+
+exports.fetchCoordFromCrossStreets = async address => {
+  googleGeoClient = googleGeoClient || new Client({});
+  cache = cache || makeCache({ maxAge: CACHE_AGE });
```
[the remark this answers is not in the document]
piratefsh: oh good catch. i meant to remove the cacheInstance in makeCache. This way, it is created and stored in `fetchCoordFromCrossStreets.js`. That should allow for varying cache instances, wdyt?

src/lib/airtable/tables/requests.js
piratefsh: good point. i thought about it but was wary because i didn't have a way to test `findOpenRequestsForSlack`. refactoring out the two helper functions seemed safe enough. if you or someone can help me test it, we can go down the path you suggested.

src/lib/cache/makeCache.js
```diff
@@ -0,0 +1,17 @@
+const LRU = require("lru-cache");
```
piratefsh, line 1: @alexquick updated `makeCache` so it doesn't store an instance!

src/api/delivery-needed/fetchCoordFromCrossStreets.js
mjmaurer: sweet. no worries if we go a little over the limit. and +1 to Alex's point

src/api/delivery-needed/fetchCoordFromCrossStreets.js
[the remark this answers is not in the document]
mjmaurer: I like the makeCache per client! looks good

src/api/delivery-needed/fetchCoordFromCrossStreets.js
```diff
@@ -0,0 +1,47 @@
...
+  if (cache.peek(cacheKey)) {
+    geoResult = cache.get(cacheKey);
+  } else {
+    // if not found, fetch from Google and cache
+    geoResult = await googleGeoClient.geocode({
+      params: {
+        address,
+        region: "us",
+        components: {
+          locality: "New York City"
```
mjmaurer, line 31: Should we just make this Brooklyn as well? Not sure why I made it NYC in the first place. It doesn't preclude results from other places

piratefsh: maybe to not preclude any other borough addresses? I'll try changing this to BK and see how it works out. we can always tweak/parameterize if needed.

piratefsh: im looking at the geocode api and it seems like there are other ways to restrict accuracy, there's sublocality and neighborhood. i'm going to keep that for a future PR as it's a different concern for now.

PR #79, opened 2020-05-22
Display requests marked for driving clusters on /delivery-needed map

piratefsh:
## Summary
- Add driving clusters as a separate layer on map. This will make it easier to customize or filter in the future without affecting the regular requests
- Add a checkbox toggle to show/hide delivery clusters. False by default.
- Had to refactor ClusterMapLayers to accommodate this.
- Refactored RequestPopup to display a Chip for a driving cluster request + add household size

Jul-04-2020 20-07-50
Screen Shot 2020-07-04 at 8 04 52 PM

## Testing
Visit http://localhost:3000/delivery-needed with prod data, you should see red and 

## Known issues
- Colors chosen are not very a11y friendly.
- Two popups may be open at the same time, one from each layer. This is because each layer handles its own popup. Possible to fix this by sharing some state for the popup. Shouldn't be too distracting for now, will open issue.


src/api/delivery-needed/index.js
```diff
@@ -12,67 +77,25 @@ exports.deliveryNeededRequestHandler = async (req, res) => {
...
-      const timestamp = metaJSON.slack_ts;
-      slackPermalink = await slackapi.chat.getPermalink({
-        channel,
-        message_ts: timestamp
-      });
-    } catch {
-      console.error(`[deliveryNeededRequestHandler] could not fetch slack URL
-        for requestCode: ${r.fields[code]} channel: ${metaJSON.slack_channel} and timestamp: ${metaJSON.slack_ts}`);
+  return res.send({
+    requests: {
```
piratefsh, line 92: send requests as separate FeatureCollections. this means the whole payload is no longer valid geoJSON but these child elements are. could not find way to include multiple FeatureCollections in a geoJSON.

src/lib/airtable/tables/requests.js
```diff
@@ -83,10 +83,7 @@ exports.findDeliveryNeededRequests = async () => {
       })
       .all();
 
-    return [
-      requests.filter(r => !requestNotInSlack(r)).filter(notForDrivingCluster),
-      null
-    ];
+    return [requests.filter(r => !requestNotInSlack(r)), null];
```
piratefsh, line 86: don't filter out driving cluster. TODO: also include driving cluster but *not* in slack

src/webapp/components/ClusterMap.js
```diff
@@ -93,21 +95,40 @@ const ClusterMap = ({ geoJsonData, containerStyle = {} }) => {
...
+          id="drivingClusterRequestsSource"
+          geoJsonSource={{
+            type: "geojson",
+            data: drivingClusterRequests,
+            cluster: true,
+            clusterMaxZoom: 14,
+            clusterRadius: 30
+          }}
+        />
+        <ClusterMapLayers
```
piratefsh, line 122: two layers now

src/webapp/components/ClusterMapLayers.jsx
```diff
@@ -36,35 +34,39 @@ const makePopupData = (features, lngLat) => {
   });
 };
 
-const ClusterMapLayers = ({ geoJsonData, paramRequest }) => {
+const ClusterMapLayers = ({ geoJsonData, paramRequest, sourceId, color }) => {
```
piratefsh, line 37: refactor to customize each layer with different data source and color

src/webapp/components/RequestPopup.js
```diff
@@ -43,34 +52,60 @@ const RequestPopup = ({ requests, closePopup }) => {
...
-            </Link>
-          ) : (
-            <Typography variant="body2">
+          <Typography variant="body2">
+            {str("webapp:deliveryNeeded.popup.requestCode", {
+              defaultValue: `Request code:`
+            })}
+            {meta.Code}
+          </Typography>
+
```
piratefsh, line 81: add chips for household size and driving cluster

src/api/delivery-needed/index.js
```diff
@@ -3,85 +3,105 @@ const { fields } = require("~airtable/tables/requests");
...
+  forDrivingClusters,
+  householdSize
+} = fields;
 
-  const requestsWithCoordsPromises = requestObj.map(async r => {
-    let metaJSON = {};
-    let slackPermalink = {};
-    const location = await fetchCoordFromCrossStreets(
-      `
+const makeFeature = async r => {
```
piratefsh, line 17: This refactor was mainly to pull out logic making geojson feature objects.

src/api/delivery-needed/index.js
```diff
@@ -3,85 +3,105 @@ const { fields } = require("~airtable/tables/requests");
...
+    properties: {
+      title: r.fields[code],
+      meta: {
+        [code]: r.fields[code],
+        [crossStreetFirst]: r.fields[crossStreetFirst],
+        [crossStreetSecond]: r.fields[crossStreetSecond],
+        [neighborhoodAreaSeeMap]: r.fields[neighborhoodAreaSeeMap],
+        [firstName]: r.fields[firstName],
+        [forDrivingClusters]: Boolean(r.fields[forDrivingClusters]),
+        [householdSize]: r.fields[householdSize],
```
piratefsh, line 70: added household size properties. this will be useful for driving clusters.

src/api/delivery-needed/index.js
```diff
@@ -3,85 +3,105 @@ const { fields } = require("~airtable/tables/requests");
...
+  const regularRequestsPromises = requestObj
+    .filter(r => !r.fields[forDrivingClusters])
+    .map(makeFeature);
+
+  const clusterRequestsPromises = requestObj
+    .filter(r => Boolean(r.fields[forDrivingClusters]))
+    .map(makeFeature);
+
+  const regularRequests = await Promise.all(regularRequestsPromises);
+  const clusterRequests = await Promise.all(clusterRequestsPromises);
```
piratefsh, line 95: fetch regular requests and cluster request separately so we can divide them up in different layers on the map.

src/webapp/components/ClusterMap.js
```diff
@@ -26,78 +28,33 @@ const makeBounds = geoJsonData => {
   return bounds;
 };
 
-const getRequestParam = () => {
```
piratefsh, line 29: refactor helper fn and alert components out as this file was getting unwieldy.

src/lib/airtable/tables/requests.js
mjmaurer: this filtering was also to ensure create delivery request still works when >8 requests are assigned. I can follow up with a different issue tho!

piratefsh: can you say more about `ensure create delivery request still works when >8 requests are assigned.`? in what way is it related to filtering out `notForDrivingCluster`?

mjmaurer: ah my mistake. this method is for requests that are already posted.

Issue #76, opened 2020-05-22
Display informational message on page when there are no deliveries to display


piratefsh:
## What
Display an info alert banner in top of ClusterMap when there are no deliveries available to display.
https://github.com/crownheightsaid/mutual-aid-app/blob/master/src/webapp/components/ClusterMap.js#L38

Example of an alert banner atop map:
![Screen Shot 2020-05-15 at 5 36 08 PM](https://user-images.githubusercontent.com/1868400/82704112-e14e4880-9c42-11ea-9930-b425edcbd290.png)

## Dependencies
To avoid merge conflict, allow this PR to be merged first https://github.com/crownheightsaid/mutual-aid-app/pull/71 as it introduces the alert banner.

## Why
Slack context: https://crownheightsmutualaid.slack.com/archives/C010AUQ6DFD/p1590158599155100
We do not display deliveries that aren't posted on Slack yet or are marked for driving clusters. Sometimes this displays an empty map and is confusing for volunteers

PR #80, opened 2020-05-27
Add alert banner when no requests are found.

sandfort:
# Summary
Display an alert banner on `/delivery-needed` if no requests are found. Delivery requests that haven't been posted to Slack yet or are marked for driving clusters don't show up on this map, so this alert should clear up some confusion. This PR should close issue #76 
maa-pr-80

### DIFF
--- a/src/webapp/components/ClusterMap.js
+++ b/src/webapp/components/ClusterMap.js
@@ -96,7 +96,7 @@ const ClusterMap = ({ geoJsonData, containerStyle = {} }) => {
   // there is a requestCode but the request object does not exist
   const paramRequestNotFound = requestCode && !paramRequest;
 
-  const noRequestsFound = geoJsonData.length === 0;
+  const noRequestsFound = geoJsonData.features.length === 0;
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


src/webapp/components/ClusterMap.js
```diff
@@ -85,12 +96,18 @@ const ClusterMap = ({ geoJsonData, containerStyle = {} }) => {
   // there is a requestCode but the request object does not exist
   const paramRequestNotFound = requestCode && !paramRequest;
 
+  const noRequestsFound = geoJsonData.length === 0;
```
piratefsh, line 99: This should check for `geoJsonData.features.length`. `geoJsonData` is the GeoJSON FeatureCollection object, which has a property `features` array that contains the markers. 

For reference, this is what gets returned from the /delivery-needed/requests endpoint: https://crownheightsma.herokuapp.com/api/delivery-needed/requests.json

sandfort:
@piratefsh Thanks for the feedback! I made the code change you requested and updated the description. It doesn't look like I can actually edit this PR's "linked issues", probably because I'm not a collaborator on this repo, so you may have to do that. I linked the issue in the description, though. Let me know if there's anything else I should change!

mjmaurer:
yup! very sorry for the long delay

PR #82, opened 2020-06-01
adding a minimal view to neighborhood finder for use in embeds

### LINKED ISSUE #81: Create /neighborhood-finder view without header for Airtable block use
## Problem
The Neighborhood Finder map https://crownheightsma.herokuapp.com/neighborhood-finder is used in the Airtable as a block, which appears as below:

**Initial view**
Screen Shot 2020-05-29 at 3 55 15 PM

Users will have to scroll in that block to see the address field and block. 

**Scrolled down:**
Screen Shot 2020-05-29 at 3 55 23 PM

## Suggested solution
Can we create a URL param e.g. ?minimal_view=true that makes the address field and map visible in the Airtable block, and moves it to below the map? For intake volunteers, I don't think it's super important for the title and desc to be visible if they are already familiar with the tool.

## Permissions  needed
- access to intake Airtable, or just create a block with the map in staging Airtable. it should show up the same either way.

eldilibra:
Fixes #81 

See #81 for summary and screengrabs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "mocha": "^7.1.1",
     "module-alias": "2.2.2",
     "morgan": "1.10.0",
+    "query-string": "^6.12.1",
     "react": "^16.13.0",
     "react-dom": "^16.13.0",
     "react-helmet": "^5.2.1",
--- a/src/webapp/pages/NeighborhoodFinder.js
+++ b/src/webapp/pages/NeighborhoodFinder.js
@@ -13,11 +13,15 @@ import { useTranslation } from "react-i18next";
 import QuadrantMap from "../components/QuadrantMap";
 import SaveNeighborhoodDataInput from "../components/SaveNeighborhoodDataInput";
 import sharedStylesFn from "../style/sharedStyles";
+import queryString from "query-string";
 
 const useStyles = makeStyles(theme => ({
   ...sharedStylesFn(theme),
   mapRoot: {
     flex: 1
+  },
+  rootMinimalView: {
+    flexDirection: "column"
   },
   formRoot: {
     flex: 1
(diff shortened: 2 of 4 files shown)


src/webapp/pages/NeighborhoodFinder.js
```diff
@@ -85,8 +91,9 @@ export default function NeighborhoodFinder() {
   };
 
   return (
-    <Box className={classes.root}>
+    <Box className={classes.root + (minimalView ? " " + classes.rootMinimalView : "")}>
```
piratefsh, line 94: you can use string templates here! e.g. `${classes.root} ${minimalView ? classes.rootMinimalView  : '""}`

package.json
```diff
@@ -59,6 +59,7 @@
     "mocha": "^7.1.1",
     "module-alias": "2.2.2",
     "morgan": "1.10.0",
+    "query-string": "^6.12.1",
```
piratefsh, line 62: good call using a lib here, there are places in the codebase where i was parsing this manually, we can replace that with query-string now.

src/webapp/pages/NeighborhoodFinder.js
eldilibra: I'll make this change :+1: . Note that there will be an extraneous space now, though.

piratefsh: oh actually we should be using the classnames lib here. maybe for a future PR? but i think it's an acceptable thing to have extra whitespace since functionality is affected.

eldilibra: Updated!

eldilibra: +1 for classnames, +1 for a future PR :~). The extra space doesn't affect a thing here, so no worries at all.

piratefsh:
@eldilibra thanks for your work! so next steps, historically @mjmaurer does deploys, i think we want it to be less blocking in the future, but in the meantime, I'm just tagging you here.

mjmaurer:
I'd love for someone to be well versed in deploying! It's really easy, but it causes downtime for 10 minutes. When we get around to the all-hands, I'll ask who's interested in learning to deploy

I also chose heroku for other groups to hit the ground running, but maybe it's time for DigitalOcean

PR #85, opened 2020-06-05
Add streets URL to slack app post

### LINKED ISSUE #78: Link to delivery-needed map from Slack app message
## What
Screen Shot 2020-05-22 at 3 36 50 PM

Cross street info on Slack request post (as highlighted above) should link to `https://crownheightsma.herokuapp.com/delivery-needed?request=<REQUEST CODE>`

## Why
This makes it easier for volunteers to quickly look up the location of the request + makes the map easier to get to.

## Dependencies
This should happen after https://github.com/crownheightsaid/mutual-aid-app/issues/77 is completed as cluster deliveries get posted to Slack but are not available on the map right now.

piratefsh:
## Summary
Pretty straightforward change to link out to Delivery Needed map from the Slack request post.
It'll link out to something like `https://crownheightsma.herokuapp.com/delivery-needed?request=FDNNBQ`.

Example from my dev Slack app:
Screen Shot 2020-06-05 at 6 19 27 PM

Hover on link:
Screen Shot 2020-06-05 at 6 19 33 PM

Closes #78 

## Open questions
The map link causes an attachment to appear in the post. Is there a way to tell Slack not to add attachments for the post? I tried looking up ways to do this but couldn't find anything. We can block attachments for any URL domain, but unsure if that's the best way forward. 

Slack post attachment I'm referring to:
Screen Shot 2020-06-05 at 6 12 51 PM

### DIFF
--- a/src/lib/strings/locales/en/slackapp.json
+++ b/src/lib/strings/locales/en/slackapp.json
@@ -133,7 +133,9 @@
           "name": "Need"
         },
         "streets": {
-          "name": "Cross Streets"
+          "name": "Cross Streets",
+          "mapUrl": "https://crownheightsma.herokuapp.com/delivery-needed?request={{requestCode}}",
+          "streetsWithMapUrl": "<{{- mapUrl}}|{{streets}}>"
         },
         "language": {
           "name": "Language"
--- a/src/slackapp/flows/createDeliveryRequest.js
+++ b/src/slackapp/flows/createDeliveryRequest.js
@@ -367,14 +367,17 @@ function suggestedTemplate(payload, request) {
     .filter(s => s !== undefined)
     .join(" & ");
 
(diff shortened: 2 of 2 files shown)


mjmaurer:
Thanks! There is an option for [the postMessage we make](https://github.com/crownheightsaid/mutual-aid-app/blob/master/src/slackapp/flows/createDeliveryRequest.js#L157) to "unfurl" media or not. 

[Docs here](https://api.slack.com/methods/chat.postMessage). Think we want  both `unfurl_links` and `unfurl_media` set to false

piratefsh:
@mjmaurer excellent, exactly what i was looking for! see screenshot for currents state without unfurled media:

Screen Shot 2020-06-06 at 11 48 50 AM

PR #99, opened 2020-06-19
mapbox token had wrong name

ksullivan2:
The token was named incorrectly in the example file, which led to a "missing mapbox token" error and a lot of troubleshooting. =)


mjmaurer:
Thanks!

PR #96, opened 2020-06-15
[NEW ENV VAR] Add twilio to neighborhood map 

jorganisak:
#18 

- Adds button next to "Email Aid Resource Links" on Neighborhood Finder page called "SMS Aid Resource Links"
- Moves both buttons to new component called `SendResourcesButtons`
- Created new component called `SendSmsDialog` - contains validation for phone number input
- Body of SMS is editable within `SendSmsDialog`, user inputs phone number
- Adds endpoint at `POST /api/send-sms` with payload `{smsMessage, phoneNumber}`
- Endpoint will 404 if `TWILIO_AUTH_TOKEN`, `TWILIO_PHONE_NUMBER`, `TWILIO_SID` are not present in env vars

sms_resources_demo

PR #103, opened 2020-06-23
added twilio serverless functions! for phone line + sms 

mab253:
inside src/twilio: 

- functions/pause.js runs when someone calls the hotline (plays an .mp3 voicemail from /assets, asks for dialtone input, and then redirects to sms/pauses-resources if the caller requests text messages)
- functions/sms/pause-resources.js runs when someone texts the hotline, explaining that the phone line is on pause and offering resources (2 text messages)
- assets folder includes voicemail message about pausing the phone, mp3

functions are initiated and deployed via the [twilio serverless toolkit](https://github.com/twilio-labs/serverless-toolkit)

currently working on the dev number, deployed @ https://mutual-aid-4526-dev.twil.io


src/twilio/assets/index.html
```diff
@@ -0,0 +1,86 @@
+<!DOCTYPE html>
```
mjmaurer, line 1: do we need these html and css assets? it doesn't look like they're served anywhere

couldn't / shouldn't we store assets like pause1.mp3 here?

src/twilio/functions/sms/pause-resources.js
```diff
@@ -0,0 +1,23 @@
...
+    })
+    .then(function() {
+      twilioClient.messages.create({
+          to: event.From,
+          from: event.To,
+          body:
+            "In the meantime we can provide other resources in Brooklyn. We will have a list on our website at crownheightsmutualaid.com. We will post updates on facebook https://www.facebook.com/groups/496603171016990 and instagram, @crownheightsaid.\n\nYou can visit https://mutualaid.nyc/resources-groups to find a Mutual Aid group in your neighborhood. You can also text FOOD to 726879 to make a reservation to pick up food at a pantry within walking distance of your home, in any borough.\n\nStay safe!"
```
mjmaurer, line 14: since both these messages might be used for #96, i'd love to find a way to put these in the /lib/strings directory without requiring all the dependencies. 

what if you created a `twilio.json` file in `src/lib/strings/locales/en` that is the same format as the other ones? You can import json as an object in node, so it should be pretty nice except the ugly import which will look something like:
`const twilioStrings = require(../../../lib/strings/locales/en/twilio.json`

index.js
```diff
@@ -56,7 +56,7 @@ app.use(bodyParser.urlencoded({ extended: true }));
 // ==================================================================
 
 if (process.env.TWILIO_AUTH_TOKEN) {
-  app.post(
+  /*app.post(
```
mjmaurer, line 59: I think we can delete this twilio section depending on the answer to the package.json comment.

if we do delete it, maybe just make a note here that twilio is deployed separately and so written as a subpackage

package.json
```diff
@@ -91,7 +91,8 @@
     "i18next-scanner": "2.11.0",
     "ngrok": "3.2.7",
     "prettier": "^1.18.2",
-    "release-it": "^13.6.2"
+    "release-it": "^13.6.2",
+    "twilio-run": "^2.7.0"
```
mjmaurer, line 95: it would be awesome to have an npm script in this root package.json to run twilio locally like the other apps (so `local:twilio`).

it wouldn't have to run the express index.js, since twilio-run supports ngrok and the like. we'd just have to point twilio-run to the src/twilio directory

src/twilio/assets/index.html
mab253: ya the html + css in there are just the defaults i think, i will delete them!

and pause1.mp3 is stored in src/twilio/assets ... it has a path that doesn't show /assets, but that's where the file is!

package.json
mab253: cool good idea!
(i made it default to running ngrok, because that's really the only way to test via a dev number ... does that make sense?)

src/twilio/functions/sms/pause-resources.js
jorganisak: Added this in #96 !
https://github.com/crownheightsaid/mutual-aid-app/pull/96/files#diff-ff63639130e2c980af1103809f45ce69R14

mab253: cool thnx joe!

index.js
mab253: looks like @jorganisak also took this out in his PR

mab253: alright, deleted!

mab253:
i think this 1 is ready for review! :alien: :robot:

src/twilio/functions/sms/pause-resources.js
```diff
@@ -0,0 +1,25 @@
+exports.handler = function firstSms(context, event, callback) {
+  const twilioClient = context.getTwilioClient();
+  twilioClient.messages
+    .create({
+      to: event.From,
+      from: event.To,
+      body:
+        "You have reached Crown Heights Mutual Aid. We are a network of neighbors helping neighbors. We are volunteers who give our time, pool our resources, and stand in solidarity with our neighbors in need.\n\nBetween June 21st and July 19th 2020 we will not be able to take new calls. We are using this time to call back the thousands of neighbors who called us before June 21st.\n\nIf you called us before June 21st, your call is recorded. We are prioritizing our Crown Heights neighbors at this time, but will do our best to call everyone back. We are planning to re-open our phone line on July 19th 2020."
+    })
+    .then(function seecondSms() {
```
mjmaurer, line 10: nit: seeeeeecond :)

mab253: haha nice catch ... fixed this ✅

Issue #88, opened 2020-06-09
Add table view for delivery needed map


piratefsh:
## What
Add a table/list view of requests to https://crownheightsma.herokuapp.com/delivery-needed with the same information available on Slack.

## Why
Delivery Needed Map (DNM) only has a map view right now. This was intended to augment the slack app requests so volunteers can browse for open requests by location.

It would be better to also have a table view of requests so volunteers can do all the browsing on the DNM page. The popovers only show minimal information (first name, x-streets and request code).

They still have to sign up for the request on Slack, but hopefully this makes less back and forth.

## Considerations
- I don't have designs yet, but if you are up for it, you can draft something up
- Hovering/clicking on marker on the map should highlight the table/list entry somehow. And vice versa. Kind of like how the [yelp map view](https://www.yelp.com/search?find_desc=cookies&find_loc=Ditmas+Park%2C+Brooklyn%2C+NY&ns=1) works.
- Add ability to filter and search could be a second iteration feature.
- We average about 30-50 requests so there is no pagination at the moment. Might wanna consider what might happen if we have more requests.

allthesignals:
I'm happy to work on this one!

mjmaurer:
Moving this comment from the PR to the issue:

> I have a bigger ask that would be SUPER useful. would it be possible to sort open requests by time since the slack post? (and also display the lapsed time)
> 
> The message ts field which is used as a thread ID is also a Unix timestamp (https://www.unixtimestamp.com/index.php), so there wouldn't need to be any extra network calls
> 
> This would be a good supplement to #63

piratefsh:
Closed because this was merged

Issue #63, opened 2020-04-28
Script for sending unclaimed grocery delivery requests to slack


mjmaurer:
This is something Bed Stuy Strong does. Right now, volunteers in CHMA do this manually. Ideally it would just look like the post below. No need to implement scheduling, it should only be a `/scripts` script. Scheduling can be added in heroku.

![Slack   delivery_volunteers   Bed-Stuy Strong   8 new items](https://user-images.githubusercontent.com/3537081/80547406-1b9b3180-8986-11ea-863c-b6275932510f.png)

PR #104, opened 2020-06-23
88 table view

allthesignals:
Addresses #88 

Adds a table to display deliveries along with hyperlink. Does not paginate, but prevents table row overflow.

image

Next steps are to hook up the map so that rows highlight on hover.


src/webapp/components/DeliveryTable.js
```diff
@@ -0,0 +1,58 @@
...
+const DeliveryTable = ({ rows }) => {
+  const classes = useStyles();
+
+  return (
+    <TableContainer className={classes.container} component={Paper}>
+      <Table aria-label="simple table">
+        <TableHead>
+          <TableRow>
+            <TableCell>Code</TableCell>
+            <TableCell align="right">Cross Street #1</TableCell>
```
piratefsh, line 26: We should be using `str` translations here as well. See README for details on how to add i18n.

mjmaurer: These strings in particular might already exist in the webapp namespace!

allthesignals:
> I have some comments but overall this looks dope! It's going to be really useful as we see increased deliveries from the temporary phone line closure to clear our queue.
> 
> Per the ticket #88, can you also add more fields as posted in #community_needs? The more fleshed out the table is, the easier it is for volunteers to decide what they can pick up. This includes household size, language, description, etc. which are particularly important for deciding how big a job will be.
> 
> All of this data isn't in the payload for the map but should be pretty easy to add on from the API. This can happen in a separate PR as well if you want to keep this one small!
> 
> Thank you for your contribution!

Thanks, I'll address all of these ASAP!

mjmaurer:
I have a bigger ask that would be SUPER useful. would it be possible to sort open requests by time since the slack post? (and also display the lapsed time)

The message `ts` field which is used as a thread ID is also a Unix timestamp (https://www.unixtimestamp.com/index.php), so there wouldn't need to be any extra network calls

allthesignals:
> I have a bigger ask that would be SUPER useful. would it be possible to sort open requests by time since the slack post? (and also display the lapsed time)
> 
> The message `ts` field which is used as a thread ID is also a Unix timestamp (https://www.unixtimestamp.com/index.php), so there wouldn't need to be any extra network calls
> 
> This would be a good supplement to #63

You bet! I think I can work on this tomorrow if that's alright. I realize the table needs a lot more work before it'll be useful.

piratefsh:
good idea re: sorting! i also think its ok to ship a minimum version like what we have right now and iterate on it. keeps PRs small and easier to review as well.

piratefsh:
as long as we communicate with volunteers whats going on, we should be good to take an iterative approach.

mjmaurer:
Yup feel free to do a different PR. Does this PR have an associated issue that I'm missing?

allthesignals:
https://github.com/crownheightsaid/mutual-aid-app/issues/88

allthesignals:
@mjmaurer I'm currently out of town dealing with a family matter — if someone wants to finish this before I am able to get back, that's totally fine

mjmaurer:
any eta? I might be able to take this up and I'll try to make sure the commits remain yours

piratefsh:
In the interest of moving forward with this feature, imma propose that we address existing issues from comments above, assess if it is safe and helpful to go to prod and iterate from there. wdyt? 

thanks @allthesignals for starting the work. I'm happy to take over this PR from here if no one else wants to.

mjmaurer:
makes sense to me!

allthesignals:
Sorry @mjmaurer @piratefsh , I'm still away dealing with some issues. I could look at it late next week at the earliest. I have no problem with anyone who wants to take it over.

allthesignals:
Looking at this again. Rebased with changes, a few merge conflicts. 

To sort by slack post, I'll need to figure out how to get that data in.

allthesignals:
Oh woops let me get the str changes in

src/webapp/components/DeliveryTable.js
```diff
@@ -0,0 +1,85 @@
...
+            <TableCell align="right">
+              {str("webapp:zoneFinder.label.timestamp")}
+            </TableCell>
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          {formattedRows.map((row) => (
+            <TableRow key={row.Code}>
+              <TableCell component="th" scope="row">
+                {row.Code}
```
mjmaurer, line 62: can you replace these row field strings with those in lib/airtable/tables/requests?

allthesignals: @mjmaurer sure! But how do I import _~airtable/tables/requests_ into the react app?

If I follow, we're using these row field strings to avoid weird property names ("Cross Street #1").

`import Requests from "~airtable/tables/requests";`  yields

> Module not found: Can't resolve '~airtable/tables/requests' in '/Users/wmattgardner/mutual-aid-app/src/webapp/components'

Importing with relative path throws errors about the dependencies of the thing being imported.

Any ideas? Am I totally missing something? Thanks!

piratefsh: hey @alllthesignals! Michael is out atm, Sherminn here. I think we probably don't have path aliasing set up for the react app, so you might just have to use relative path e.g. `../../lib/airtable`. Since this is non-blocking, plus we have other places in the webapp where we need to fix this, im going to go ahead and merge this first and follow up with a new issue to handle hardcoding of airtable field keys. 

This table view will be important for the upcoming work for https://github.com/crownheightsaid/mutual-aid-app/issues/122, where volunteers can pick up delivery straight from the map.  thank you for your contribution!

Issue #90, opened 2020-06-10
Add alerting for critical failures


mjmaurer:
I think in the form of a post to wg_tech makes the most sense.

Following cases should alert:
- [ ] Failure during airtable polling.

PR #108, opened 2020-07-03
Send error notification to slack

jorganisak:
~Needs https://github.com/crownheightsaid/airtable-change-detector/pull/14 to work~

https://github.com/crownheightsaid/airtable-change-detector/pull/14 enhances current errFunc api to also catch errors within the passed function from the caller of `pollWithInterval`, but not necessary for this to work

- Passes errFunc as arg to `pollWithInterval`
- Sends `error.stack` to `wg_tech`
- Doesn't send if the same error has been sent within 2 hours (internal memory, error history won't survive app restart)

#90


mjmaurer:
action item here is to move to the current errorFunc api

jorganisak:
@mjmaurer - I think this uses the current api, (passing errFunc as 4th arg to `pollWithInterval`) But I could be missing something..

jorganisak:
Updated PR description for clarity

src/workers/airtable-sync/paymentWorker.js
```diff
@@ -9,12 +9,17 @@ const {
...
 } = require("~airtable/tables/paymentRequests");
+const sendErrorNotification = require("~slack/errorNotification");
 const newExternalDonorPayment = require("./actions/payments/newExternalDonorPayment");
 const newPaymentRequest = require("./actions/payments/newPaymentRequest");
 const updateReimbursementMessage = require("./actions/payments/updateReimbursementStatus");
 
 const defaultInterval = 10000;
 
+const errFunc = (error) => {
+  sendErrorNotification(error);
```
mjmaurer, line 20: optional nit: can you make the function async and await the sendError? functionally the same, but a bit more resistant to a potential weird future change-detector update

PR #109, opened 2020-07-03
Linter/prettier upgrade fixes

jorganisak:
Looks like prettier version was bumped, and linting is complaining. Fortunately all was well after a `npm run fix`

Not sure if we are opinionated about the prettier recommended defaults... from what I can tell all of these changes are either for
- requiring trailing commas on lists
- requiring parenthesis around single args in functions


mjmaurer:
Thanks so much. This one is on me, didn't realize prettier had a major version update. I ran `npm update` for the newest release, and thought all the dependencies were fixed to minor or patch. 

As far as our style opinions, feel free to object. I brought over the style config from another of my nodejs / react projects.

PR #117, opened 2020-07-09
Add hover text to household size chip

### LINKED ISSUE #116: Add hover text for household size indicator
Screen Shot 2020-07-08 at 1 32 53 PM

The pink chip indicates household size but doesn't have any alt text right now. For usability, we should add hover text that says something like `Household size of 2`.

Location of source: https://github.com/crownheightsaid/mutual-aid-app/blob/master/src/webapp/components/RequestPopup.js#L83

saoirse-zee:
This adds a tooltip to the "household size" chip on the deliveries map.

Screen Shot 2020-07-09 at 11 57 13 AM

This can be seen in staging. Check out the delivery at Dean St and Troy Ave.

Resolves #116

### Questions

- Should the text include the number of people? E.g. "Household size: 17"? Or is it meant more of an accessibility aid to the icon?

- Is it clear what the tooltip refers to? There are options to add arrows, etc. I can play around with that if this is unclear.

- I see that some text is translated, other is not. It seems that labels, such as the tooltip I added, are not translated. Am I reading that right?

- I committed the changes to the lockfile, as this was the practice in past projects. Should I be doing this?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -60,7 +60,11 @@
           "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-1.2.13.tgz",
           "integrity": "sha512-oWb1Z6mkHIskLzEJ/XWX0srkpkTQ7vaopMQkyaEIoq0fmtFVxOthb8cCxeT+p3ynTdkk/RZwbgG4brR5BeWECw==",
           "dev": true,
-          "optional": true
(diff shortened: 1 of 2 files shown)


mjmaurer:
@saoirse-zee feel free to merge!

PR #120, opened 2020-07-21
Twilio functions update: the phone lines are open!

mab253:
3 new functions: 2 for voice calls (play intro mp3 and gather input; then re-direct and write to Airtable), and 1 for SMS

(these functions are already deployed on mutual-aid-4526-dev.twil.io as they needed to happen over the weekend, but this PR is to update the code so it matches)


mab253:
also ... i am not sure why the npm run lint check failed ... it worked locally ... ?

sandfort:
> also ... i am not sure why the npm run lint check failed ... it worked locally ... ?

Looks like there's a lot of lint changes in files that you didn't otherwise make changes to. Something like this happened to me before, I think it's because you ran the linter with a different rule set. When it happened to me, I rolled back the commit with the lint fixes and ran `npm install` again (that may have been incidental) and it worked.

mab253:
thank you @sandfort and @mjmaurer ! finally able to fix this - it was the linter following old rules.

i also made another functions update - only requests self-identified as "crown heights" on the phone will write to the Airtable.

src/twilio/functions/after-input.js
```diff
@@ -0,0 +1,72 @@
...
+  let manyc = "";
+  let carcluster = "";
+
+  const Airtable = require('airtable'); // eslint-disable-line
+  Airtable.configure({
+    endpointUrl: "https://api.airtable.com",
+    apiKey: context.AIRTABLE_API_KEY // eslint-disable-line
+  });
+
+  const base = Airtable.base("apppK7mrvMPcwtv6d");
```
sandfort, line 18: I think this airtable functionality (and the interactions below in this file) already exist in our library: https://github.com/crownheightsaid/mutual-aid-app/blob/master/src/lib/airtable/tables/requests.js

mab253: that's true - ! these little phone functions in src/twilio are what we deploy separately to [Twilio Runtime](https://www.twilio.com/docs/runtime), so they are serverless - unless i'm missing something, we don't want to pull from the rest of app here, we want these to live on their own 

(but we list them here for transparency/version control etc.!)

sandfort: Oh I see! That makes sense, I didn't realize these were meant to be independent.

Issue #102, opened 2020-06-22
<= instead of < for detecting reimbursement complete


mjmaurer:
https://crownheightsmutualaid.slack.com/archives/C011D4YBWHL/p1592853085112900?thread_ts=1592853085.112900&cid=C011D4YBWHL

sandfort:
It looks like we're already using `<=` for this: https://github.com/crownheightsaid/mutual-aid-app/blob/9385eaa099074ee42de4dc0c7ee5b97157603780/src/slackapp/event-listeners/reimbursementReply.js#L57
Might it be a floating point issue?

mjmaurer:
Yup! You're probably correct.

PR #121, opened 2020-07-24
Test reimbursement reply

sandfort:
This PR adds a set of unit tests for the Slack event listener for reimbursement replies. It is meant to address issue #102 , although it doesn't contain any substantive changes to the code in question. I wasn't able to reproduce this particular issue and I'm still not sure what caused it. At least that piece of code is tested now, and this should also provide a model for testing similar pieces code.


mjmaurer:
thanks! can you speak to the schema refactor? did it make it more testable?

sandfort:
> thanks! can you speak to the schema refactor? did it make it more testable?

Yep! That's pretty much it. Specifically, I'd like to be able to mock the behavior of some of our Airtable functions without also mocking the constants, and Jest's module mocking mocks the entire module. So, separate modules. I also just think it's a nice clean separation of responsibilities.

src/slackapp/event-listeners/reimbursementReply.js
```diff
@@ -18,17 +19,31 @@ const filterAndReply = async (event) => {
...
+
+  let reimbursementChannel;
+
+  try {
+    reimbursementChannel = await findChannelByName(REIMBURSEMENT_CHANNEL);
+    if (!isReimbursementReply(event, reimbursementChannel.id)) {
+      return;
+    }
+  } catch (e) {
+    console.error("Error finding reimbursement channel", e);
```
piratefsh, line 31: quick question, non blocking for PR: do console.error logs show up on our application runtime logs on Heroku?

sandfort: I'm actually not sure... I used console.error here because I saw it elsewhere in the codebase, didn't think about it much beyond that.

mjmaurer: they do! and they stand out a bit more

Issue #125, opened 2020-08-01
Clicking on request in table view should highlight request on map and vice versa


piratefsh:
## Summary
Table view of requests has been [recently added](https://github.com/crownheightsaid/mutual-aid-app/pull/104/). However, there is no correlation between the map and table view at the moment. We want to populate the table view with more data e.g. the grocery list in the future. To make the table view more useful, focus on a request on the map should highlight the request table row as well. Conversely, clicking/hovering on the request on the table should open the popover for the request on the map.

## Requirements
- If the table row is out of view, it should scroll into visible view
- Similarly, if the map popover is out of view, it should be panned into view
- This may require sharing state across components i.e. ClusterMap and DeliveryTable. _May_ have to introduce some sort of state management like Redux.

![Screen Shot 2020-08-01 at 1 46 04 PM](https://user-images.githubusercontent.com/1868400/89107331-4ae69200-d3fe-11ea-850f-c50afa4c8d08.png)
(rough sketch, please use a subtle highlight background color)

piratefsh:
closed by https://github.com/crownheightsaid/mutual-aid-app/pull/137

PR #137, opened 2020-08-11
Clicking on request on map highlights request on table

### LINKED ISSUE #127: Display DaysOpenChip in DeliveryTable
This is a pretty straightforward one!
## Summary 

Screen Shot 2020-08-01 at 2 13 21 PM
^ this is the DaysOpenChip

Display [DaysOpenChip](https://github.com/crownheightsaid/mutual-aid-app/blob/05759ac828934e62bac1d643e3711ecb3482641b/src/webapp/components/DaysOpenChip.js#L8) in [DeliveryTable](https://github.com/crownheightsaid/mutual-aid-app/blob/05759ac828934e62bac1d643e3711ecb3482641b/src/webapp/components/DeliveryTable.js#L13) in a new column. This will make it easy for volunteers to browse for requests that have been open for a long time.

piratefsh:
... and many other improvements

Closes #127 and #125 
## Summary
Screen Shot 2020-08-10 at 8 27 38 PM

This builds on @allthesignals work that introduced the DeliveryTable, ty!

- Display table view beside map view
- Clicking on table row will trigger popup on map. Conversely, clicking on marker on map will highlight table row
- Add DaysOpenChip and HouseholdSize and DrivingClusterChip to table
- Moved around table columns to better reflect importance of information

## Requirements fulfilled
- Both table and map should reflect driving clusters when `Show driving clusters` is checked
- If table row is not in view, it should scroll to be in view when selected and vice versa if marker is not visible on map

## Testing
Checkout this branch and click around the map and table. Comments, questions, clarifications welcome!


.eslintrc.js
```diff
@@ -27,7 +27,8 @@ module.exports = {
     "react/jsx-wrap-multilines": ["off"],
     "react/destructuring-assignment": ["off"],
     "react-hooks/rules-of-hooks": ["error"],
-    "radix": ["off"] //fixed in ES5
+    "radix": ["off"], //fixed in ES5
+    "import/prefer-default-export": ["off"]
```
piratefsh, line 31: most of our `webapp/helpers` export multiple functions, imo we don't have to enforce this.

src/webapp/components/ClusterMap.js
```diff
@@ -103,12 +94,14 @@ const ClusterMap = ({ geoJsonData, containerStyle = {} }) => {
           sourceId="requestsSource"
           paramRequest={paramRequest}
           color="orangered"
+          data={reqFeatures}
```
piratefsh, line 97: need to pass in data to layers now so they can look up the focused request.

src/webapp/components/ClusterMapLayers.jsx
```diff
@@ -3,58 +3,77 @@
...
-      } = paramRequest;
-      setPopup(makePopupData([paramRequest], coordinates));
+    // display popup if request code is present in URL search param
+    // but only on first render
+    if (paramRequest && initialRender) {
+      setPopup(makePopupData([paramRequest], paramRequest.geometry.coordinates));
+      setInitialRender(false)
+    }
+
+    //find focused request and display popup
```
piratefsh, line 59: when `focusedRequest` is present (when set from table), trigger popup

src/webapp/components/ClusterMapLayers.jsx
```diff
@@ -78,10 +97,10 @@ const ClusterMapLayers = ({ geoJsonData, paramRequest, sourceId, color }) => {
                 5,
                 30,
                 20,
-                40
-              ]
+                40,
```
piratefsh, line 100: just some formatting here

src/webapp/components/DrivingClusterChip.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
```
piratefsh, line 1: refactored out from `RequestPopup`

src/webapp/components/HouseholdSizeChip.js
```diff
@@ -0,0 +1,30 @@
+import React from "react";
```
piratefsh, line 1: also refactored out from `RequestPopup`

src/webapp/context/ClusterMapContext.js
```diff
@@ -0,0 +1,8 @@
+import { createContext } from "react";
```
piratefsh, line 1: new context to handle which request is currently being focused

src/webapp/helpers/time.js
```diff
@@ -0,0 +1,8 @@
+import { differenceInDays, fromUnixTime } from "date-fns";
```
piratefsh, line 1: refactored out from `RequestPopup`.

src/webapp/components/ClusterMapLayers.jsx
```diff
@@ -3,58 +3,77 @@
...
+          properties: {
+            meta: { Code },
+          },
+        }) => Code === focusedRequestId
+      );
+    if(focusedRequest){
+      setPopup(makePopupData([focusedRequest], focusedRequest.geometry.coordinates))
     }
-  }, []);
+  }, [focusedRequestId]);
```
mjmaurer, line 70: not related to this PR, but surprised we don't need paramRequest here as well. wouldn't an in-app navigation with the code param reuse the same paramRequest otherwise?

src/webapp/components/ClusterMap.js
piratefsh: yeah its a little messy and ideally there'll be a single store for requests 😅

src/webapp/components/ClusterMapLayers.jsx
piratefsh: thats a good question. so unfortunately, the way i set up paramRequest before was that it only worked on the first map render, so updating the url param wouldn't re-focus the request. though your comm ent made me think that we should update the url param on focus. it would be easier for me to remove the old paramRequest logic and replace handling url params with `focusedRequest` instead.

piratefsh:
added a few commits to merge with master as well as fix a bug where RequestPopup was not surfacing multiple requests at the same location

src/webapp/components/DeliveryTable.js
```diff
@@ -1,79 +1,126 @@
...
+              onClick={() => setFocusedRequestId(row.Code)}
+            >
+              <TableCell>
+                <DaysOpenChip
+                  timeOnly
+                  daysOpen={daysSinceSlackMessage(row.slackTs)}
+                />
+              </TableCell>
+              <TableCell>
+                {row["First Name"]}
```
ifearcompilererrors, line 107: should this say something like "Status" or "About"?

src/webapp/components/DeliveryTable.js
```diff
@@ -1,79 +1,126 @@
...
+                <Box className={classes.chipRow}>
+                  {row["For Driving Clusters"] && <DrivingClusterChip />}
+                  <HouseholdSizeChip size={row["Household Size"]} />
```
ifearcompilererrors, line 110: should we only show the household chip if it has a size value other than n/a?

src/webapp/components/DrivingClusterChip.js
ifearcompilererrors: heads up although im sure you're aware, we mayyy have merge conflicts- happy to work em out with you!

src/webapp/components/DeliveryTable.js
piratefsh: good q. im displaying n/a just to keep it obvious that this data wasn't available as this is what is displayed on the Slack post

src/webapp/components/DeliveryTable.js
piratefsh: good call, i'm gonna rename this to `Neighbor`!

PR #140, opened 2020-08-12
Display all delivery requests by default

### LINKED ISSUE #139: Display all delivery requests by default
## Context
Request from cluster coordination: 

> hey tech! As we've moved away from fulfilling BCH tickets, the cluster coordination team (tbh kinda only me but I got support from one other person lol) was wondering if we could have the map by default show all delivery needed requests, including anything that says "yes" in the "For Driving Clusters" column, and then having the checkbox filter it to only show the "yes" in "For Driving Clusters" tix (this second part is not necessary but if it's easy then would be cool to include!). The idea is that when we do intake for a large family, that would become a "For Driving Clusters" ticket, and then we would encourage the #cars_and_bikes folks to claim those tickets using the filter on the map. LMK if that makes sense! not a high urgency project. thanks for all you do!

## Summary
Right now we only have a single option to `Show Driving Clusters`. Instead of just having one check box, proposing to replace this with to two checkboxes:

- [ ] Driving Cluster Requests
- [ ] Regular requests

And both should be checked by default. This should allow for easy filtering for all circumstances in the future.

saoirse-zee:
- Replaces the single checkbox ("Show Driving Clusters") with two checkboxes: "Driving Cluster Requests" and "Regular requests".
- Both are checked by default.
- Updates some names for the sake of consistency.

Resolves #139

Here's what it looks like with the change:

demo-map-checkboxes

I considered stacking the checkboxes vertically, but decided against it since it pushed the map further down the page. Here's what that would look like: 

map-checkboxes-vertical

### DIFF
--- a/src/webapp/components/ClusterMap.js
+++ b/src/webapp/components/ClusterMap.js
@@ -31,13 +31,14 @@ const makeBounds = (features) => {
 
 const ClusterMap = ({ geoJsonData, containerStyle = {} }) => {
   const requestCode = getRequestParam();
-  const [showDrivingClusters, setShowDrivingClusters] = useState(false);
+  const [showDrivingRequests, setShowDrivingRequests] = useState(true);
+  const [showRegularRequests, setShowRegularRequests] = useState(true);
 
   let paramRequest;
   const { requests, drivingClusterRequests } = geoJsonData;
   const { features: reqFeatures } = requests;
   const { features: clusterFeatures } = drivingClusterRequests;
-  const allRequests = showDrivingClusters
+  const allRequests = showDrivingRequests
     ? [...reqFeatures, ...clusterFeatures]
     : reqFeatures;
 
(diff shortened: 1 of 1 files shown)


src/webapp/components/ClusterMap.js
```diff
@@ -31,13 +31,14 @@ const makeBounds = (features) => {
...
-  const [showDrivingClusters, setShowDrivingClusters] = useState(false);
+  const [showDrivingRequests, setShowDrivingRequests] = useState(true);
+  const [showRegularRequests, setShowRegularRequests] = useState(true);
 
   let paramRequest;
   const { requests, drivingClusterRequests } = geoJsonData;
   const { features: reqFeatures } = requests;
   const { features: clusterFeatures } = drivingClusterRequests;
-  const allRequests = showDrivingClusters
+  const allRequests = showDrivingRequests
```
saoirse-zee, line 41: I didn't mess with this logic, and I think that's okay, but wanted to flag it in case I'm missing something.

piratefsh:
i think its a good call to rename, `driving clusters` is kind of confusing alongside map clusters

PR #142, opened 2020-08-19
[138] Add Delivery Table fields - Need, Timeline, & Description

### LINKED ISSUE #138: [SMS Pickup] Add remaining fields to DeliveryTable
Add all fields that appear in the Slack post to DeliveryTable:

-Timeline
-Need
-Description

Parts:
- Add those 3 fields to the delivery-needed/requests.json endpoint:https://github.com/crownheightsaid/mutual-aid-app/blob/cf4ad4938c412b42143e69ffb19bf9e72efe4baf/src/api/delivery-needed/index.js#L63
- Display fields on DeliveryTable only (no need to show in RequestPopup).
  - for Timeline: clarify that it's the timeline set by neighbor and is distinct from DaysOpen
  - for Description:  potentially several lines long, figure out way to display and expand elegantly in table without causing it to be too long

mirandawang:
Fixes #138 

Add fields from the Slack form to the Delivery Requests Table:
- Need
- Timeline
- Description

Also a little bit of refactoring: moved the Delivery Table Row into its own component and rendering the header row in a function. 

Open to suggestions for ordering of these fields! I tried to match how they are currently ordered in the Slack form.

I decided to label Timeline as "Requested Delivery Time" to convey the difference between its meaning & "Days Open". What do you think? Open to suggestions for this too!

Since Description can be long and the table was already getting pretty wide with the two extra fields, I decided to use the Collapsible Table material UI component. I think it looks pretty nice! I also added a Tool Tip when you hover over the arrow icon which will say show or hide description.

The description will expand whenever the row is in focus. Check it out!

Aug-19-2020 16-21-43

**To test these changes:**

1. Pull this branch
2. Make sure you're set up to pull from the staging Heroku instance
3. Load `/delivery-requests` page
4. You should see the new fields added and description expanded when you click on a row!

(side note: idk why my git account isn't properly set up to these commits on this computer :sob:)


src/api/delivery-needed/index.js
```diff
@@ -70,6 +73,9 @@ const makeFeature = async (r) => {
         [firstName]: r.fields[firstName],
         [forDrivingClusters]: Boolean(r.fields[forDrivingClusters]),
         [householdSize]: r.fields[householdSize],
+        [timeline]: r.fields["Timeline"],
+        [need]: r.fields["Need"],
+        [description]: r.fields["Description"],
```
mirandawang, line 78: Is there a better way to reference these values?

piratefsh: ah actually these fields should be defined in  `fields` in `~airtable/tables/request`. you can see on line 6 of this file where I am destructuring them out from `fields`. We do it this way because fields like "Timeline" are actually Airtable column names, and if this changes, it will be easier to change them in just one place.

You can see which fields this data is being referenced from in the source where the Slack posts are constructed. So "Timeline" is actually `timeSensitivity`: https://github.com/crownheightsaid/mutual-aid-app/blob/4f56609960e8dc3e7843ace7c6818651e92bbf5a/src/slackapp/flows/createDeliveryRequest.js#L406

src/lib/airtable/tables/requests.js
```diff
@@ -322,6 +322,9 @@ const fields = (exports.fields = {
   neighborhood: "Neighborhood MA-NYC",
   householdSize: "Household Size",
   forDrivingClusters: "For Driving Clusters",
+  timeline: "Timeline",
```
piratefsh, line 325: Ah, I should have clarified before, these fields actually aren't 1:1 on naming. "Timeline" is actually timeSensitivity, "Needs" is determined in logic [here](https://github.com/crownheightsaid/mutual-aid-app/blob/4f56609960e8dc3e7843ace7c6818651e92bbf5a/src/slackapp/flows/createDeliveryRequest.js#L397) and "Description" is `notes`. Those are just labels applied for the Slack post. So there isn't a need to add new fields.

src/webapp/components/DeliveryTable.js
```diff
@@ -66,6 +51,46 @@ const DeliveryTable = ({ rows }) => {
...
+        <TableCell>
+          {str("webapp:deliveryNeeded.table.headers.crossStreets", {
+            defaultValue: "Cross streets",
+          })}
+        </TableCell>
+      </TableRow>
+    );
+  }
+
+  const renderTableBody = () => {
```
piratefsh, line 88: this can be a stateless component, e.g. `const TableRows ({formattedRows}) => {...}` and you can render it below as `<TableRows />`, which is a more recent React pattern and has performance optimizations.

piratefsh: you can also define it outside of `DeliveryTable` , it will be slightly more performant for re-renders since it doesn't redefine the function.

src/webapp/components/DeliveryTableRow.js
```diff
@@ -0,0 +1,117 @@
+import React, { useContext } from "react";
```
piratefsh, line 1: Nice refactor out!

src/webapp/helpers/getDeliveryRequestNeedFormatted.js
```diff
@@ -0,0 +1,17 @@
+const { isEqual } = require("lodash");
```
piratefsh, line 1: did you mean to remove this duplicate file?

src/lib/airtable/deliveryrequests/getDeliveryRequestNeedFormatted.js
```diff
@@ -0,0 +1,18 @@
+/**
```
piratefsh, line 1: good call refactoring it out 👍

src/webapp/components/DeliveryTable.js
```diff
@@ -66,66 +48,46 @@ const DeliveryTable = ({ rows }) => {
     };
   });
 
+  const renderTableHeadRow = () => {
```
piratefsh, line 51: nit: can you make this a stateless component too? e.g. `const TableHeadRows = ()=> .<TableRow>...</TableRow>` it can stay in this file.  keeping things consistent as components makes code easier to read!

src/webapp/helpers/getDeliveryRequestNeedFormatted.js
mirandawang: whoops, yes lol

PR #144, opened 2020-08-25
Test donor payments table

sandfort:
This PR adds some simple unit tests for the donor payments Airtable functions. It has one positive and one negative test for each of the two functions, although no tests for exceptional behavior. I hope the way I mocked the Airtable library will prove to be a good example for testing these library functions.

### DIFF
--- a/src/lib/airtable/tables/donorPayments.test.js
+++ b/src/lib/airtable/tables/donorPayments.test.js
@@ -1,0 +1,26 @@
+const mockCreate = jest.fn().mockResolvedValue('value')
+
+jest.mock("~airtable/bases", () => ({
+  paymentsAirbase: () => {
+    return {create: mockCreate}
+  }
+}))
+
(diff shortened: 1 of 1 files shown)


src/lib/airtable/tables/donorPayments.test.js
```diff
@@ -0,0 +1,92 @@
...
+      expect(result).toEqual(["value", null]);
+    });
+  });
+
+  describe("on an unsuccessful Airtable call", () => {
+    beforeEach(() => {
+      mockCreate.mockRejectedValue("error");
+    });
+
+    test("it creates a donor payment record with the given fields", async () => {
```
piratefsh, line 47: did you mean to change this to describe the error return value?

sandfort: Good catch! Copy-paste error.

PR #149, opened 2020-09-08
Test request table

sandfort:
This PR separates the Requests table schema from its functions, and also adds unit tests for those functions.


saoirse-zee:
Just noting that this work will conflict with this outstanding PR: https://github.com/crownheightsaid/mutual-aid-app/pull/153

This branch is much more substantial, so I think we should merge this one first. Tap me when this is merged and I can go back and make the changes needed to the other branch to align with the new schema. 

Otherwise no comment other than yay tests!

src/lib/airtable/tables/requests.js
```diff
@@ -224,108 +224,3 @@ const removeSlackMeta = (meta) =>
     .omit("slack_channel")
     .thru(JSON.stringify)
     .value();
-
```
piratefsh, line 227: I like refactoring this out and I think it makes a lot of sense! However, removing `fields` from exports will break a bunch of APIs and UI that are importing these fields from `requests.js`. e.g. https://github.com/sandfort/mutual-aid-app/blob/a090a31523b4d69733cf5bddbbabe27e869dfa08/src/api/delivery-needed/index.js#L10

Can I suggest exporting it to keep this support for now? we can just do `exports.fields = fields`

src/lib/airtable/tables/requests.test.js
```diff
@@ -0,0 +1,483 @@
+const mockDestroyFn = jest.fn();
```
piratefsh, line 1: These looks so good! I think this approach makes sense and will definitely help prevent any avoidable breakages in the future.

Just one thing, and this doesn't have to be fixed in this PR but I just want to bring to your attention:

I ran test coverage (`npx jest src/lib --coverage`) and it looks like most of the code is tested, but we are missing tests for a number of error cases

For reference, the `--coverage` output:

```
--------------------------|---------|----------|---------|---------|-----------------------------------
File                      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
--------------------------|---------|----------|---------|---------|-----------------------------------
All files                 |   74.01 |    68.75 |      80 |   73.86 |
 donorPayments.js         |   86.67 |      100 |     100 |   86.67 | 34-35
 donorPaymentsSchema.js   |     100 |      100 |     100 |     100 |
 paymentRequests.js       |   21.28 |        0 |   16.67 |   21.28 | 11-35,40-55,60-71,76-79,84-98,106
 paymentRequestsSchema.js |     100 |      100 |     100 |     100 |
 requests.js              |   92.86 |    96.88 |     100 |   92.78 | 49-50,89,115,149,153,212
 requestsSchema.js        |     100 |      100 |     100 |     100 |
--------------------------|---------|----------|---------|---------|-----------------------------------
```

^ most of the uncovered lines for requests.js are error cases.

src/lib/airtable/tables/requestsSchema.js
```diff
@@ -0,0 +1,106 @@
+const { airbase } = require("~airtable/bases");
```
piratefsh, line 1: Love that this is its own file now!

piratefsh (changes requested):
Hi John thanks so much for these tests! It mostly lgtm, left a couple of comments on testing error cases + `fields` export. While looking at this PR, a couple of things came to my attention as well:

- Can we move *.tests.js to a `__tests__` directory in the same directory as the file being tested? This will make it easier to navigate the files. e.g.:
 ` src/lib/airtable/tables/__tests__/requests.test.js `. By [default](https://jestjs.io/docs/en/configuration#testmatch-arraystring), Jest will be able to pick up tests in `__tests__` directories
- We can replace "test" in `package.json` with `jest`! I don't think we use `react-scripts` for testing at the moment. This way we can just run `npm run test` to get all the sweet, sweet test running!

^ These two are non-blocking for this PR, but just want to get your thoughts. If they sound like good ideas, we can make GH issues for these.

sandfort:
>     * Can we move *.tests.js to a `__tests__` directory in the same directory as the file being tested? This will make it easier to navigate the files.
Done!

>     * We can replace "test" in `package.json` with `jest`! I don't think we use `react-scripts` for testing at the moment. This way we can just run `npm run test` to get all the sweet, sweet test running!
Not sure I fully understand this one: tests can currently be run with `npm test` or `npm run test`, but those don't work if I make this change. You can see `npm test` working as-is in the CI run for previous commits.

src/lib/airtable/tables/requests.js
sandfort: Whoops! I'm actually still exporting `fields`, just doing it from `requestsSchema.js` now instead. I thought I had changed the various imports throughout the app, but I see I missed some...

src/lib/airtable/tables/requests.test.js
sandfort: This is a good point, the only reason I didn't shoot for 100% coverage is that this PR was taking long enough and I didn't want it to be too monumental.

src/lib/airtable/tables/requests.js
piratefsh: gotcha, i see you updated all the places where its imported, that works too!

.github/workflows/nodejs.yml
```diff
@@ -24,4 +24,5 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
     - run: npm ci
-    - run: npm run lint
\ No newline at end of file
+    - run: npm run lint
+    - run: npm test
```
piratefsh, line 28: nice!

piratefsh (approved):
Latest changes lgtm! thanks for taking the feedback. lets ship these tests 💪 

Re: `npm test`, my bad! i got confused because it wasn't running any tests when I ran it on my machine. Seems like by default it runs tests pertaining to latest change only when on dev:

```
$~/D/mutual-aid-app (test-request-table) $ npm test

No tests found related to files changed since last commit.
Press `a` to run all tests, or run Jest with `--watchAll`.

Watch Usage
 › Press a to run all tests.
 › Press f to run only failed tests.
 › Press q to quit watch mode.
 › Press p to filter by a filename regex pattern.
 › Press t to filter by a test name regex pattern.
 › Press Enter to trigger a test run.
```
 
I don't think we need to fix this for now as it runs all tests on CI successfully. The GH action setup is looking good 👍

PR #152, opened 2020-09-21
add sms delivery/intake/pickup and sms pod twilio functions

mab253:
additions to src/twilio/functions/sms

these functions are already deployed to Twilio serverless environment:

https://mutual-aid-4526-dev.twil.io/sms/pod-verify
(SMS bot that allows pod members to generate Airtable "intake" codes, in order to make repeated deliveries to the same neighbors and get reimbursed via CHMA system)

https://mutual-aid-4526-dev.twil.io/sms/test-delivery
(staging endpoint, for claiming deliveries via SMS)

https://mutual-aid-4526-dev.twil.io/sms/intake-delivery-txts
(prod endpoint, for claiming deliveries and communication between intake/delivery vols, via SMS)

### DIFF
--- a/src/twilio/functions/sms/intake-delivery-txts.js
+++ b/src/twilio/functions/sms/intake-delivery-txts.js
@@ -1,0 +1,99 @@
+exports.handler = function codeSms(context, event, callback) { // eslint-disable-line
+  const Airtable = require("airtable"); // eslint-disable-line
+  Airtable.configure({
+    endpointUrl: "https://api.airtable.com",
+    apiKey: context.AIRTABLE_API_KEY // eslint-disable-line
+  });
+
+  const base = Airtable.base("apppK7mrvMPcwtv6d"); // intake + vols base
+  let code = event.body.requestCode;
+  const deliveryPhone = event.body.deliveryPhone; // eslint-disable-line
+  let deliveryName = event.body.deliveryName; // eslint-disable-line
+  let intakePhone = event.body.intakePhone; // eslint-disable-line
+  let intakeName = event.body.intakeName; // eslint-disable-line
+  let body = JSON.stringify(event.body); // eslint-disable-line
+  let missinginfo = false;
+  let missingintake = false;
+  code = code.toUpperCase().trim();
+
+  console.log("incoming message received");
+  console.log(`delivery phone: ${deliveryPhone}`);
+
+  // below: looks up code in request Airtable, gets all important values about the requesting neighbor
+  base("Requests")
+    .select({
+      view: "Grid view", // from staging base
+      maxRecords: 1,
+      filterByFormula: `({Code} = '${code}')`,
+    })
+    .firstPage(function seekRecords(err, records) {
+      records.forEach(function getRecordInfo(record) {
+        console.log("Retrieved", record.get("Code"));
+        console.log("ID", record.id);
+
+        const phone = record.get("Phone");
+        let firstName = record.get("First Name");
+        if (typeof firstName === "undefined") {
+          firstName = "your neighbor";
+        }
+        const list = record.get("Intake General Notes");
+        if (typeof list === "undefined") {
+          missinginfo = true;
+        }
+        const street1 = record.get("Cross Street #1");
+        if (typeof street1 === "undefined") {
+          missinginfo = true;
+        }
+        const street2 = record.get("Cross Street #2");
+        if (typeof street2 === "undefined") {
+          missinginfo = true;
+        }
+        if (intakePhone === " ") {
+          missingintake = true;
+        }
+
+        if (typeof intakeName === "undefined") {
+          intakeName = "ready to help";
+        }
+
+        if (typeof deliveryName === "undefined") {
+          deliveryName = "a CHMA volunteer";
+        }
+
+        let deliveryText = `Thanks for taking on this delivery for ${firstName}!\nCODE = ${code}.\n\nTheir phone is ${phone}, you will need to get in touch with them about the full address. Their cross streets are ${street1} & ${street2}.\n\n${firstName}'s grocery list is: ${list}\n\nThe intake volunteer for this request is ${intakeName}. Their phone # is ${intakePhone}, and they can help if you have any questions - they'll reach out to you to follow up and make sure the delivery goes well!`;
+
+        if (missinginfo === true) {
+          deliveryText = `Thanks for taking on this delivery for ${firstName}!\nCODE = ${code}.\n\nIt looks like some important info might be missing - please follow up with #intake_volunteers on Slack, or text your intake volunteer ${intakeName} at ${intakePhone} - we'll get it sorted out!`;
+        }
+
+        if (missingintake === true) {
+          deliveryText = `Thanks for taking on this delivery for ${firstName}!\nCODE = ${code}.\n\nIt looks like some important info might be missing - please follow up with #intake_volunteers on Slack - we'll get it sorted out`;
+        }
+
+        // sends SMS out to delivery volunteer
+        const twilioClient = context.getTwilioClient();
+
+        twilioClient.messages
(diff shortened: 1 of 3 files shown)


src/twilio/functions/sms/intake-delivery-txts.js
```diff
@@ -0,0 +1,99 @@
+exports.handler = function codeSms(context, event, callback) { // eslint-disable-line
+  const Airtable = require("airtable"); // eslint-disable-line
+  Airtable.configure({
+    endpointUrl: "https://api.airtable.com",
+    apiKey: context.AIRTABLE_API_KEY // eslint-disable-line
+  });
+
+  const base = Airtable.base("apppK7mrvMPcwtv6d"); // intake + vols base
+  let code = event.body.requestCode;
+  const deliveryPhone = event.body.deliveryPhone; // eslint-disable-line
```
piratefsh, line 10: you can destructure each key like this:
```
let {
  deliveryPhone,
  deliveryName,
  intakePhone,
  intakeName } = event.body;
```
and now those variables will exist in scope and can be used. This should remove the eslint complaint as well.

src/twilio/functions/sms/intake-delivery-txts.js
```diff
@@ -0,0 +1,99 @@
...
+          intakeName = "ready to help";
+        }
+
+        if (typeof deliveryName === "undefined") {
+          deliveryName = "a CHMA volunteer";
+        }
+
+        let deliveryText = `Thanks for taking on this delivery for ${firstName}!\nCODE = ${code}.\n\nTheir phone is ${phone}, you will need to get in touch with them about the full address. Their cross streets are ${street1} & ${street2}.\n\n${firstName}'s grocery list is: ${list}\n\nThe intake volunteer for this request is ${intakeName}. Their phone # is ${intakePhone}, and they can help if you have any questions - they'll reach out to you to follow up and make sure the delivery goes well!`;
+
+        if (missinginfo === true) {
```
piratefsh, line 65: You should be able to just do `id(missinginfo)` and it should run the if block if the value is true here!

src/twilio/functions/sms/intake-delivery-txts.js
```diff
@@ -0,0 +1,99 @@
...
+  });
+
+  const base = Airtable.base("apppK7mrvMPcwtv6d"); // intake + vols base
+  let code = event.body.requestCode;
+  const deliveryPhone = event.body.deliveryPhone; // eslint-disable-line
+  let deliveryName = event.body.deliveryName; // eslint-disable-line
+  let intakePhone = event.body.intakePhone; // eslint-disable-line
+  let intakeName = event.body.intakeName; // eslint-disable-line
+  let body = JSON.stringify(event.body); // eslint-disable-line
+  let missinginfo = false;
```
piratefsh, line 15: just a nit but usually in JS variables should be camelCased

src/twilio/functions/sms/test-delivery.js
```diff
@@ -0,0 +1,102 @@
+// this endpoint version receives delivery/intake name + phone numbers from the React web app
```
piratefsh, line 1: This seems to be a duplicate of intake-delivery-txts, i think you can use the same code as intake-delivery-txts, but replace any staging values with a context variable, like how `context.AIRTABLE_API_KEY` is passed in. I don't know much about how env vars are passed in Twilio, but happy to figure it out together if you'd like!

Having less duplicate code will help reduce work and mistakes in the future when we need to update this code!

piratefsh (changes requested):
Left some feedback! Thanks for working on this Madeline!! Please lmk you need more clarification on any of the comments.

src/twilio/functions/sms/intake-delivery-txts.js
mab253: cool, made this change!

src/twilio/functions/sms/intake-delivery-txts.js
mab253: 🤖✅

src/twilio/functions/sms/intake-delivery-txts.js
mab253: 🤖✅

src/twilio/functions/sms/test-delivery.js
mab253: didn't get to this yet, but i think i know what you mean: we can have 1 twilio function, so just the 1 endpoint URL that would work for both staging and prod, that switches only which Airtable base we read. 

what's my `if` statement, then - how do I determine if it's staging? is it the heroku URL, do I ask where the request is coming from?

thnx!

piratefsh: ah i meant that it can be same code, but still two different twilio endpoints. one endpoint can be the staging and the other prod, but you deploy the same code, but set different `context.AIRTABLE_BASE`. is that possible for twilio? 

if not, we can keep two different files, but maybe abstract out the main logic. so prod and staging  call the same helper function. e.g.:

sendDeliveryTexts.js
```
function sendDeliveryTexts(airtableBase){
  // set given airtable base here
  ...
}

```

intake-delivery-texts-prod.js
```
sendDeliveryTexts(<prod airtable base>)
```

intake-delivery-texts-staging.js
```
sendDeliveryTexts(<staging airtable base>)
```

either way is fine, as long as we can minimize duplicate code!

also, just a nit, but wondering if we can use the same name for the endpoint, just with prod/staging? e.g. /intake-delivery-texts-prod and /intake-delivery-texts-staging. it will be easier for us to keep track of which is which this way!

mab253: ah i see!

yeah so how twilio creates the endpoints is with each file.  i realize they're totally duplicate but since they're stateless functions, it only reads from what's in that 1 file ... i think the only way to avoid totally duplicating would actually to have 3 files:

intake-delivery-texts-prod -> sets the base variable for prod, then calls /send-intake-delivery-texts endpoint
intake-delivery-texts-staging -> sets the base variable for staging, then calls /send-intake-delivery-texts endpoint
send-intake-delivery-texts -> knows the base from the firing event, but has the meat of the code

does that make sense? if it's ok to duplicate the code to have 2 endpoints only, i think that might be easier ..!

piratefsh: Oh I see, is it not possible to import a helper function with `require`? 

Since `require('airtable') works`, i wonder if you can just do 

```
const sendDeliveryTexts = require('./sendDeliveryTexts')
```

 where `sendDeliveryTexts.js` is the helper function file with main logic. 

Or did you have to do something special to enable require for Airtable? I'm not super familiar with stateless functions, let me know if that's not possible! Just throwing suggestions out there.

mab253: hmmmm i would have to test this! it's a good idea, i've just never tried ... 
[i know that the require("airtable") is an npm package, but that doesn't necessarily mean i couldn't require other scripts that are hosted in the twilio environment!]

not sure i could get to this today ... but probably by monday/tuesday?!

piratefsh: yeah no worries! i think we are ok to merge without that change, can you make a GH issue for this refactor so it gets tracked? looking forward to having this in the repo!!

piratefsh (approved):
lgtm, with a follow up to removing duplicated code as discussed in https://github.com/crownheightsaid/mutual-aid-app/pull/152#discussion_r495473942

src/twilio/functions/sms/test-delivery.js
mab253: yassssss post-merge refactor makes sense, we can dig into how to make it elegant

piratefsh:
@mab253 imma merge this for you since it has been approved!

PR #153, opened 2020-09-22
Replace hardcoded keys with the centralized list of Airtable keys

### LINKED ISSUE #124: Use airtable request fields in place of hardcoded keys
## Summary
We have a number of keys used to access Airtable objects that are hardcoded in the webapp. Example: https://github.com/crownheightsaid/mutual-aid-app/blob/b02423c45d11d51e6faed958a8b47e469cb0aae9/src/webapp/components/RequestPopup.js#L62

We should replace this hardcoded keys with the centralized list of Airtable keys here: https://github.com/crownheightsaid/mutual-aid-app/blob/b02423c45d11d51e6faed958a8b47e469cb0aae9/src/lib/airtable/tables/requests.js#L234

This should be importable from the webapp. 

## Bonus
Get ~airtable alias working within the webapp. I believe it's only possible to import via relative paths at the moment. This issue was faced here: https://github.com/crownheightsaid/mutual-aid-app/pull/104/files#r459213347

## Why
This is important to prevent inconsistencies as well as typos that may lead to bugs! This app is also used by other mutual aids and having these keys in a centralized place makes it easy for them to adapt this app for their airtable, much like with string translations.

saoirse-zee:
Resolves #124

### DIFF
--- a/src/lib/airtable/tables/requests.js
+++ b/src/lib/airtable/tables/requests.js
@@ -1,6 +1,6 @@
 const { merge } = require("lodash");
 const _ = require("lodash");
-const { airbase } = require("~airtable/bases");
+const { airbase } = require("../bases");
 
 const requestNotInSlack = (r) => {
   const meta = r.get(fields.meta);
--- a/src/webapp/components/RequestPopup.js
+++ b/src/webapp/components/RequestPopup.js
@@ -12,6 +12,7 @@ import HouseholdSizeChip from "./HouseholdSizeChip";
 import DrivingClusterChip from "./DrivingClusterChip";
 import { daysSinceSlackMessage } from "../helpers/time";
 import ClusterMapContext from "../context/ClusterMapContext";
+import { fields } from "../../lib/airtable/tables/requests";
 
 const useStyles = makeStyles((theme) => ({
   divider: {
@@ -70,17 +71,17 @@ const RequestPopup = ({ requests, closePopup }) => {
                 underline="always"
                 target="_blank"
               >
-                {meta["First Name"] || ""}
+                {meta[fields.firstName] || ""}
               </Link>
             ) : (
-              meta["First Name"]
+              meta[fields.firstName]
             )}
           </Typography>
 
           <Typography variant="body1">
-            {meta["Cross Street #1"]}
+            {meta[fields.crossStreetFirst]}
             {" and "}
-            {meta["Cross Street #2"]}
+            {meta[fields.crossStreetSecond]}
           </Typography>
 
           <Typography variant="body2">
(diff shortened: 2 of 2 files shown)


src/lib/airtable/tables/requests.js
```diff
@@ -1,6 +1,6 @@
 const { merge } = require("lodash");
 const _ = require("lodash");
-const { airbase } = require("~airtable/bases");
+const { airbase } = require("../bases");
```
saoirse-zee, line 3: For some reason I had to make this a relative path. The tilde path threw an error.

src/webapp/components/RequestPopup.js
```diff
@@ -12,6 +12,7 @@ import HouseholdSizeChip from "./HouseholdSizeChip";
 import DrivingClusterChip from "./DrivingClusterChip";
 import { daysSinceSlackMessage } from "../helpers/time";
 import ClusterMapContext from "../context/ClusterMapContext";
+import { fields } from "../../lib/airtable/tables/requests";
```
saoirse-zee, line 15: Same here: the absolute path (with or without tilde) didn't work.

saoirse-zee (commented):
There are a few raw airtable strings in updateMessageContent.js, but they don't show up in the airtable fields translation object. Should they be added?

src/lib/airtable/tables/requests.js
mjmaurer: very weird! do you remember the error? this should be currently in prod right?

mjmaurer:
> There are a few raw airtable strings in updateMessageContent.js, but they don't show up in the airtable fields translation object. Should they be added?

yes please!

src/lib/airtable/tables/requests.js
saoirse-zee: Yeah, super weird! The version with the tilde path is in prod, which proves that it works. Maybe this is just user error on my part.

Here's the error: `Module not found: Can't resolve '~airtable/bases' in '/Users/emmasmith/projects/mutual-aid-app/src/lib/airtable/tables'`

I see this when I run `npm run local:react`. Also when I try to build for prod using `npm run build`. The Express app runs just fine. It seems that the React side of things is not respecting the `moduleAliases`. 

Clues, advice, guidance most appreciated :)

src/webapp/components/RequestPopup.js
sandfort: Possibly this is old news, but after looking over it, it looks like the `~whatever` module aliases aren't in use anywhere in the `webapp` directory... surprising! But I think that at least means that this problem isn't because of any change in this PR.

sandfort: Also, it looks like most of the other scripts in our `package.json` are requiring `module-alias/register` which I assume is what's giving other scripts the ability to resolve those aliases. However, I'm not actually sure how to do that in `local:react` since it runs `react-scripts` instead of `node`.

PR #148, opened 2020-09-02
Add reminder to Neighborhood Finder

caroqliu:
Adds the reminder text to the json field as well as the same text as the default value in the `NeighborhoodFinder` component. Fixes #146 

[Here it is in staging](https://crownheightsma-staging.herokuapp.com/neighborhood-finder) with a screenshot of Airtable:
image

PR #154, opened 2020-09-22
Color-code the map markers according to their age

### LINKED ISSUE #126: Color code map markers to how long it has been open
Might be a bit tricky for first-time contributors, unless you have experience with Mapbox, in which case, please take this on!

## Summary
Chip displaying days open was recently added: https://github.com/crownheightsaid/mutual-aid-app/pull/118

We should also color code the delivery requests marker based on how long they are open, so volunteers can quickly see which requests have been open for a long time. So if the days open chip is green, the marker should also be green. e.g.
Screen Shot 2020-08-01 at 2 00 34 PM

## Considerations
- The cluster mode of the markers. e.g. when it displays (3) should be updated to be a more neutral color to avoid competing for attention with the color coded requests.
- Code for controlling marker colors are here: https://github.com/crownheightsaid/mutual-aid-app/blob/05759ac828934e62bac1d643e3711ecb3482641b/src/webapp/components/ClusterMapLayers.jsx#L67. Check out mapbox [documentation](https://docs.mapbox.com/mapbox-gl-js/example/data-driven-circle-colors/) for more on how this works. Right now the `step` config steps based on number of requests, and not number of days it has been open, so you may have to wrangle the source data.

saoirse-zee:
This resolves #126 

The color of the markers on the map now matches the color of the "days open" chip in the table.

delivery-map

I added a dark gray outline to the map marker with the intention of making it more legible to those who have trouble seeing low contrast colors. I experimented with just making the colors darker, but I had trouble finding a dark orange distinct enough from the red.

The cluster markers are now a "neutral" dark gray, so avoid confusing them with the "urgent" red markers.

The color coding logic is in a helper that is used by both the table and the map. I included tests for the helper.


src/webapp/components/ClusterMap.js
```diff
@@ -27,6 +28,14 @@ const makeBounds = (features) => {
   return bounds;
 };
 
+const addDaysOpen = (feature) => ({
+  ...feature,
+  properties: {
+    ...feature.properties,
+    daysOpen: daysSinceSlackMessage(feature.properties.meta.timestamp),
```
saoirse-zee, line 35: This works, but I have an idea for another approach: Instead of adding a `daysOpen` property here, we could add a `markerColor` property. This could be computed using a helper that is shared between the map marker and the color chip in the table. This would have two advantages over the existing approach: 1) If we ever change the color coding logic, we only change it in one place, and 2) We get to do the logic in javascript instead of the rather weird Mapbox expression syntax.

Just a thought. If this seems worth doing, let me know :)

src/webapp/components/ClusterMapLayers.jsx
```diff
@@ -132,8 +124,12 @@ const ClusterMapLayers = ({ data, paramRequest, sourceId, color }) => {
             sourceId={sourceId}
             filter={["!", ["has", "point_count"]]}
             paint={{
-              "circle-color": color,
-              "circle-radius": 6,
+              "circle-color": ["step",
```
saoirse-zee, line 127: If we made the change I suggested above, this would be something like `"circle-color": ["get", "markerColor"]`

src/webapp/components/ClusterMap.js
piratefsh: yes, i think this sounds good, especially for reason 2)!

piratefsh (changes requested):
Thanks for this! Code looks good, including your suggestion to add an abstracted function. 

### For documentation purposes, would you mind adding:

1) link to the issue this closes
2) description of what this PR does (short and simple is OK)
3) screenshots

Example of a PR description: https://github.com/crownheightsaid/mutual-aid-app/pull/148

### Some UI feedback:
Screen Shot 2020-09-28 at 8 18 40 PM

1. seems like the clusters are now appearing as grey. with white text, it seems rather low contrast, which can be difficult for folks with low visibility to read the numbers. can you make it a darker grey? i think it makes sense to have clusters be a neutral color now that individual markers have colors
2. im worried about the `orange` markers because they will also have low contrast, and hence easy to miss. would you mind experimenting with adding a dark border around each marker, or using a darker shade of orange? you can also update the color on the DaysOpenChip to keep it consistent.

But otherwise, looking great!! Thanks for spending time figuring our mapbox!

src/webapp/helpers/map-urgency.js
```diff
@@ -0,0 +1,41 @@
+/**
```
piratefsh, line 1: Nice, i like that this is reusable now!

src/webapp/helpers/map-urgency.test.js
```diff
@@ -0,0 +1,27 @@
+const { getUrgencyStyles } = require("./map-urgency");
```
piratefsh, line 1: Tests!! I'm so happy, thanks for writing these!

piratefsh (approved):
This looks amazing!! It will definitely help highlight requests that have been open for long. Thanks so much for incorporating the feedback, refactoring and tests.

sandfort (approved):
Looks great! Love to see more testing and reusability!

Issue #158, opened 2020-10-08
Use `Date Changed to Delivery Needed` to calculate number of days a request has been open


piratefsh:
## Summary
`X days open` data on map relies on `slack_ts`. As part of SMS pickup work, we want to remove any reliability on Slack postings. Instead, we should use the timestamp in  `Date Changed to Delivery Needed` on Airtable requests.

This field currently does not exist in our static list of [Airtable fields](https://github.com/crownheightsaid/mutual-aid-app/blob/master/src/lib/airtable/tables/requests.js#L234) yet, and should be added. This field gets updated on Airtable when a request gets updated to `Delivery Needed`. 

## Where
Logic for converting slack_ts to days open here: https://github.com/crownheightsaid/mutual-aid-app/blob/master/src/webapp/helpers/time.js. This function should be updated to take in a timestamp from `Date Changed to Delivery Needed`, and its usage should be updated throughout the `webapp/` part of this repo.

Note: this should be merged after https://github.com/crownheightsaid/mutual-aid-app/pull/154 as there was some refactoring to days open logic in that PR

PR #160, opened 2020-10-09
Use 'Date Changed to Delivery Needed' field to calculate days a request has been open

jreinlein:
Fixes issue #158 

This is my first PR, so let me know if anything is amiss! 

If this looks good, a `hacktoberfest-accepted` label on the PR would be much appreciated 😁 

**Reviewers**
- Let me know if the new `invalid` time label is poorly thought-out or implemented.
- Is there a test suite to run?

## Changes
- Added `dateChangedToDeliveryNeeded` field to requests' fields
- Added `invalid` color state to handle rows that are missing the field (time function return `-1`)
- Removed `slackTs` field (no longer needed? duplicate of `timestamp`?)
## Testing
- ran `npm run lint` and `npm run test`
- Populated the `Date Changed to Delivery Needed` field for a few rows and saw it render properly in the table

**Screenshot:**
image

### DIFF
--- a/src/api/delivery-needed/index.js
+++ b/src/api/delivery-needed/index.js
@@ -41,7 +41,6 @@ const { TWILIO_SMS_DELIVERY_ENDPOINT } = process.env;
 const makeFeature = async (r) => {
   let metaJSON = {};
   let slackPermalink = {};
-  let timestamp;
 
   const location = await fetchCoordFromCrossStreets(
     `
@@ -68,7 +67,7 @@ const makeFeature = async (r) => {
 
   try {
     const channel = metaJSON.slack_channel;
-    timestamp = metaJSON.slack_ts;
+    const timestamp = metaJSON.slack_ts;
     slackPermalink = await slackapi.chat.getPermalink({
       channel,
       message_ts: timestamp,
@@ -100,7 +99,6 @@ const makeFeature = async (r) => {
         [intakeNotes]: r.fields[intakeNotes],
         need,
         slackPermalink: slackPermalink.ok ? slackPermalink.permalink : "",
-        timestamp,
         dateChangedToDeliveryNeeded: r.fields[dateChangedToDeliveryNeeded],
       },
     },
--- a/src/api/delivery-needed/index.js
+++ b/src/api/delivery-needed/index.js
@@ -33,6 +33,7 @@ const {
   timeSensitivity,
   intakeNotes,
   supportType,
+  dateChangedToDeliveryNeeded,
 } = fields;
 
 const { TWILIO_SMS_DELIVERY_ENDPOINT } = process.env;
@@ -100,7 +101,7 @@ const makeFeature = async (r) => {
         need,
         slackPermalink: slackPermalink.ok ? slackPermalink.permalink : "",
         timestamp,
-        slackTs: metaJSON.slack_ts || "",
+        dateChangedToDeliveryNeeded: r.fields[dateChangedToDeliveryNeeded],
       },
     },
   };
--- a/src/webapp/components/DeliveryTable.js
+++ b/src/webapp/components/DeliveryTable.js
@@ -5,7 +5,6 @@ import TableCell from "@material-ui/core/TableCell";
 import TableContainer from "@material-ui/core/TableContainer";
 import TableHead from "@material-ui/core/TableHead";
 import TableRow from "@material-ui/core/TableRow";
-import Tooltip from "@material-ui/core/Tooltip";
 import Paper from "@material-ui/core/Paper";
 import { makeStyles } from "@material-ui/core/styles";
 import { useTranslation } from "react-i18next";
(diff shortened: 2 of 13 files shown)


src/webapp/helpers/time.js
```diff
@@ -1,8 +1,15 @@
...
-const daysSinceSlackMessage = (slackTs) => {
-  const datePosted = fromUnixTime(Number(slackTs));
+/**
+ * @param {string} isoTimestamp - ISO timestamp, eg "2020-10-09T18:38:28.000Z"
+ * @returns {number} - difference in days since provided ISO timestamp
+ */
+const getDaysSinceIsoTimestamp = (isoTimestamp) => {
+  if (!isoTimestamp) {
+    return -1;
+  }
```
jreinlein, line 10: I'm unable to view the prod Airtable, but based on how many rows were missing the `Date Changed to Delivery Needed` field in Staging, I thought this would be a prudent safety check 🤷‍♂️

src/webapp/helpers/time.js
```diff
@@ -1,8 +1,15 @@
-import { differenceInDays, fromUnixTime } from "date-fns";
+import { differenceInDays, parseISO } from "date-fns";
 
-const daysSinceSlackMessage = (slackTs) => {
```
piratefsh, line 3: You might need to check if this function is used anywhere else in the codebase.

Also, consider writing some tests for this if you can, this will help a lot with code coverage!

src/webapp/helpers/map-urgency.js
```diff
@@ -3,6 +3,9 @@
  * usable by Material UI.
  */
 export const urgencyStyles = {
+  invalid: {
+    backgroundColor: "black",
```
piratefsh, line 7: i think the background color for this can be "grey" when this data is unavailable. this will downplay the importance of the "n/a" label.

src/api/delivery-needed/index.js
```diff
@@ -100,7 +101,7 @@ const makeFeature = async (r) => {
         need,
         slackPermalink: slackPermalink.ok ? slackPermalink.permalink : "",
         timestamp,
-        slackTs: metaJSON.slack_ts || "",
```
piratefsh, line 103: We can also remove `timestamp` here! I believe it was a duplicate added awhile ago for the DaysOpenChip. Just in case you haven't already, might be worth manuallyy checking in the codebase if `slackTs` is used anywhere else. Unfortunately a lot of code is untested, so a manual search will be the best way to go about it,

piratefsh (changes requested):
Looking real good!!! Thanks for working on this. I'll get the hacktoberfest labels for you.

Re: invalid: good call implementing some sort of fallback! However, I feel like invalid would be pretty confusing for volunteers as it is not clear if something went wrong. I think what would be best is to use "n/a".

Also, bonus UI improvement if you are up for it: what could be useful is to add a little help icon on the 'Days Open' column header explaining that this number is the number of days that this request has been in `Delivery Needed` status, and that If it says "n/a", this information is not available for some reason and they should contact #tech_support on slack. We should expect this data to always be available in prod.

We do not have many tests, but you can run them with `npm run test`, then hit `a` to run the full test suite (by default it only runs test for changes since last commit in dev).

Feel free to ask for any clarifications/suggest alternatives!

src/webapp/helpers/map-urgency.js
jreinlein: good point!

src/api/delivery-needed/index.js
jreinlein: Already did :) It's only used as a parameter name [here](https://github.com/crownheightsaid/mutual-aid-app/blob/master/src/lib/airtable/tables/requests.js#L195)

jreinlein:
> Re: invalid: good call implementing some sort of fallback! However, I feel like invalid would be pretty confusing for volunteers as it is not clear if something went wrong. I think what would be best is to use "n/a".

YES, agreed! I came to the same conclusion after mulling it over a few hours after I posted the PR.

> Also, bonus UI improvement if you are up for it: what could be useful is to add a little help icon on the 'Days Open' column header explaining that this number is the number of days that this request has been in `Delivery Needed` status, and that If it says "n/a", this information is not available for some reason and they should contact #tech_support on slack. We should expect this data to always be available in prod.

👏  love it 👏

src/webapp/components/DeliveryTable.js
```diff
@@ -5,28 +5,40 @@ import TableCell from "@material-ui/core/TableCell";
...
   return (
     <TableRow>
       <TableCell />
-      <TableCell>
-        {str("webapp:deliveryNeeded.table.headers.daysOpen", {
-          defaultValue: "Days open",
-        })}
-      </TableCell>
+      <Tooltip
+        title={str("webapp:deliveryNeeded.table.headers.daysOpen.tooltip")}
```
jreinlein, line 32: Reviewer(s): I wasn't sure if I should include a `defaultValue`. I opted _not_ to because it feels redundant, but let me know if you'd like me to.

jreinlein:
@piratefsh - I added the tooltip directly to the column header instead of a `?` button (1) to stay consistent with the other tooltip used in the table and (2) to prevent clutter if we want to add more context info to other headers

src/webapp/helpers/time.test.js
```diff
@@ -0,0 +1,23 @@
+const { getDaysSinceIsoTimestamp } = require("./time");
```
piratefsh, line 1: nice tests!

src/webapp/components/DeliveryTable.js
piratefsh: I think in general we always have a defaultValue just in case the translation files are missing the copy. It does seem pretty redundant, I agree. Originally the translation files exist so that other mutual aids forking this app have an easy way to customize copy. I am still trying to figure out if this is still the case.

Usually when I'm working on the app, I write the `str` statement with the defaultValue and when you run `npm run find-new-strings` to add the translation entry, it automatically pulls the defaultValue in the translation files. There's documentation here but it's not very findable: https://github.com/crownheightsaid/mutual-aid-app/blob/783ee668a1e8acd992179105f88804053205f485/src/lib/strings/DEVELOPING.md

Maybe to be safe we can add the defaultValue for now.

piratefsh (approved):
looks good!! just one comment re: string defaultValue, but ok to merge after its addressed!

hmm re: tooltip, since its crowding the header, perhaps it makes sense to have it on the badges themselves? it's not obvious UI to hover on the table header, but it might be on the badge. at any rate, this can be addressed in the future PR as it doesn't take away from the current UI.

thanks again for working on this!!

jreinlein:
Noted! I'll add the `defaultValue` to the strings and  move the tooltip to be on the badge (agreed that it's more intuitive)

Issue #147, opened 2020-09-01
[Neighborhood finder] Display error message on Address field if address is out if CHMA bounds


piratefsh:
*Contributing for the first time? Please see our README on how to join our community*

## Context
Follow up to https://github.com/crownheightsaid/mutual-aid-app/issues/146, please ensure that #146 is completed first as it is higher priority. 

## Summary
Instead of just showing a static message which can be easily overlooked, we should surface some sort of error message to indicate to volunteers than an address is outside of CHMA bounds. We have the GeoJSON for the neighborhood bounds, but will need a way to check if a set of coordinates is within it or not.

We can surface this error as a form validation error. Rough mockup:
<img width="1406" alt="Screen Shot 2020-09-01 at 6 24 30 PM" src="https://user-images.githubusercontent.com/1868400/91912772-ef6c2600-ec81-11ea-8b47-9828a59a6dde.png">

Considerations:
- should prevent `Update a request with the above address` if address is out of bounds?

patrikkei:
Hello, can i work on this task? :)

piratefsh:
hi @patrikkei  are you added to the CHMA Slack? if yes, please complete 'Before you start' https://github.com/crownheightsaid/mutual-aid-app/blob/master/DEVELOPING.md#Before-you-start so we can give you permission to pick up issues! if not, you can join the Slack by signing up to be a volunteer and completing an orientation session, please see our README for more details!

PR #162, opened 2020-10-16
Use material UI error and helper text to handle response error or out…

sephhh:
… of bounds address

Fixes issue #147 

Changes
- if user enters an address outside delivery area bounds, we display a red error message under the address field and mark the field red
- moves existing generic error message for a failed request to display the same way
- email button with resource links always displays (if this isn't intended behavior lmk!)

Before submitting an address:
Screen Shot 2020-10-16 at 12 24 04 PM

Generic error when request fails (i.e. no address entered, or address not found)
Screen Shot 2020-10-16 at 12 24 11 PM

Error when address found but not in bounds
Screen Shot 2020-10-16 at 12 24 30 PM

### DIFF
--- a/src/webapp/pages/NeighborhoodFinder.js
+++ b/src/webapp/pages/NeighborhoodFinder.js
@@ -15,6 +15,37 @@ import QuadrantMap from "webapp/components/QuadrantMap";
 import SaveNeighborhoodDataInput from "webapp/components/SaveNeighborhoodDataInput";
 import sharedStylesFn from "webapp/style/sharedStyles";
 
+const getError = ({ responseError, responseData, str }) => {
+  if (responseError) {
+    return (
+      <>
+        {str("webapp:zoneFinder.geoError.message")}
+        &nbsp;
+        <a
+          href={str("webapp:slack.techChannelUrl")}
+          target="_blank"
+          rel="noopener noreferrer"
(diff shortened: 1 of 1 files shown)


src/webapp/pages/NeighborhoodFinder.js
```diff
@@ -15,6 +15,37 @@ import QuadrantMap from "webapp/components/QuadrantMap";
...
+      </>
+    );
+  }
+
+  const addressIsOutsideDeliveryArea =
+    responseData && responseData.quadrant === null;
+  if (addressIsOutsideDeliveryArea) {
+    return str(
+      "webapp:zoneFinder.message.reminder",
+      `Reminder: the boundaries on this map show a defined delivery area,
```
piratefsh, line 40: to make the error message clearer, i think we can update it to `This address is outside of CHMAs delivery area. Per our Delivery Area proposal that passed 7/25. CHMA cannot make any deliveries outside of this area, no exceptions.`

piratefsh (approved):
Left a comment about clarifying some copy, but other than that, looks good to me!! Thanks so much for working on this!

Ok to merge once the copy has been addressed 👍 

Also if you are up to it, this component can use some tests. Doesn't have to be part of this PR, but it would really help us.

Issue #159, opened 2020-10-09
List more useful information on localhost:3000 


piratefsh:
## Summary
When the webapp gets started up with `npm run local:react`, we are greeted with a blank page that says `Slack attack`. This was because early on in this app the server was primarily a Slack bot. This view is not very useful for new or existing developers.

What could be more useful is a list of all webapp routes (and maybe API routes? open to suggestions) and maybe a link to developing resources. This is not a comprehensive ticket, open to suggestions!

jreinlein:
I'm curious to hear input from someone more familiar with the app and its direction: **how do we feel about adding a navigation element**? I'm not sure how often we expect users to switch between sub-apps (neighbourhood finder, delivery needed, etc?), but having a method of navigation may be helpful if the answer isn't "never" 😅 

Material UI offers a few options, but the ones that make the most sense to me (in desc order) are:
1. [Drawer](https://material-ui.com/components/drawers/)
3. [Breadcrumbs](https://material-ui.com/components/breadcrumbs/)
2. [Tabs](https://material-ui.com/components/tabs/)
4. [Bottom nav](https://material-ui.com/components/bottom-navigation/)

Of course, we could always go with Sher Minn's recommendation and provide links and some copy on the landing page to limit the scope of this piece of work

piratefsh:
completed with #163

PR #163, opened 2020-10-16
Add landing page

jreinlein:
Fixes issue #159 

I took a stab at adding useful info to the landing page (default `/` path). Let me know what you think! **Very open to suggestions.**

### Testing
- Clicked around, looked at console for errors
- `npm run test` and `npm run lint`

### [🎦  Screen recording](https://drive.google.com/file/d/121QF4KAVzDBGpw8Oo7o_dbvJggBA6KSM/view?usp=sharing)
(GitHub doesn't accept `.mov` or `.mp4` 😠  ...any suggestions for how to upload recordings?)


src/webapp/pages/Home.js
```diff
@@ -1,7 +1,98 @@
...
+  icon: {
+    marginRight: "0.5rem",
+  },
+  subAppPaperLink: {
+    fontSize: "1.5rem",
+    display: "flex",
+    alignItems: "center",
+    justifyContent: "center",
+    "&:hover": {
+      backgroundColor: "hsla(0, 0%, 0%, 0.05)",
```
piratefsh, line 25: minor nit: i think the bg color can be a _tad_ darker for better contrast on brighter screens.

piratefsh (approved):
it looks so good!! i think the way you have it organized makes sense, plus having the links really adds context. one minor comment but otherwise lgtm!

re: screengrabs, i think linking out is fine. i usually make gifs with Giphy and upload them to the PR but it does limit the length of the demo.

src/webapp/pages/Home.js
jreinlein: @piratefsh you want a darker colour when hovering over the Papers that link to the other apps? Or do you mean the actual bg colour outside of the Papers?

piratefsh: ah I meant the actual BG outside of the Paper! maybe increase the opacity on backgroundColor here. feel free to tweak to see what looks good.

jreinlein: Modifying the background color would require [affecting the root](https://stackoverflow.com/questions/48186478/custom-theme-background-color) styling which would affect each page as far as I can tell, so as a compromise I've increased the `elevation` used in the Papers

image

PR #171, opened 2020-10-28
Improve coverage for existing airtable test suites

sandfort:
This PR increases test coverage in existing airtable function library test files to 100%! You can check coverage yourself with:
`npx jest src/lib/airtable/tables/__tests__/ --coverage`

For those wondering why I'm using `npx` here, I can't actually figure out how to use the `jest` executable otherwise, even though I'm pretty sure that's what `react-scripts` is using.


dannysullivan (approved):
LGTM

piratefsh (approved):
Amazing! thank you so much! this will help with confidence when we make changes in the future, esp since lib/airtable is still heavily used!

PR #179, opened 2020-12-16
Handle airtable services field to take either string or array

piratefsh:
We are planning to change “What type(s) of support are you seeking?” field from a MULTI-SELECT to a SINGLE SELECT. This means the value is going from an array to a string. 

This patch will handle either case so things won't break when we switch over.

### DIFF
--- a/src/lib/airtable/deliveryrequests/getDeliveryRequestNeedFormatted.js
+++ b/src/lib/airtable/deliveryrequests/getDeliveryRequestNeedFormatted.js
@@ -9,13 +9,15 @@ exports.getDeliveryRequestNeedFormatted = (services) => {
   if (!services) {
     return "Not stated";
   }
-  let servicesString = ""
 
-  if (typeof services === "string"){
-    servicesString = services
-  } else if (Array.isArray(services)){
-    servicesString = services.join()
+  let servicesString = "";
+
+  if (typeof services === "string") {
+    servicesString = services;
+  } else if (Array.isArray(services)) {
+    servicesString = services.join();
   }
+
   if (servicesString === fields.supportType_options.delivery) {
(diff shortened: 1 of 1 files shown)


src/lib/airtable/deliveryrequests/getDeliveryRequestNeedFormatted.js
```diff
@@ -6,10 +6,18 @@
 const { fields } = require("~airtable/tables/requestsSchema");
 
 exports.getDeliveryRequestNeedFormatted = (services) => {
-  if (services === undefined) {
+  if (!services) {
```
piratefsh, line 9: handle any falsy case

src/lib/airtable/deliveryrequests/getDeliveryRequestNeedFormatted.js
```diff
@@ -6,10 +6,18 @@
...
     return "Not stated";
   }
-  const servicesString = services.join(", ");
+
+  let servicesString = "";
+
+  if (typeof services === "string") {
+    servicesString = services;
+  } else if (Array.isArray(services)) {
+    servicesString = services.join();
```
mab253, line 18: does it need the (", ") here for the join, like it was before? or will it know to join with comma as the separator?

piratefsh: oh good catch. hmm so it shouldn't make a difference for this case since we are only direct matching for `"Groceries / Shopping"` but i'll revert it anyway!

saoirse-zee:
This looks good to me!

mab253 (approved):
lookin' good! 🤖

PR #182, opened 2021-01-27
new strings: for Slack posts

mab253:
points to SMS map instead of suggesting that users comment on a Slack thread to pick up delivery!

[also adding new Twilio audio file, from serverless update around x-mas time]


mab253:
[please ignore how that last commit is called "major update" hahah - that was for my own forked repo!]

piratefsh (approved):
one question but otherwise lgtm!!

mab253:
> one question but otherwise lgtm!!

cool! wait i'm not seeing the question anywhere @piratefsh - happy to answer tho!

src/twilio/functions/sms/intake-delivery-txts.js
```diff
@@ -20,7 +20,6 @@ exports.handler = function codeSms(context, event, callback) { // eslint-disable
   // below: looks up code in request Airtable, gets all important values about the requesting neighbor
   base("Requests")
     .select({
-      view: "Grid view", // from staging base
```
piratefsh, line 23: @mab253 oh sorry my Q was this change intentional? just checking since this wasnt mentioned in the desc!

mab253: o yes! i believe this is already deployed via the twilio flow - but this is because in the "Requests" view that's not the Staging base, there's no view called "Grid view." And if we don't add the view parameter, it searches through the whole table 👍 

thnx for flagging!!

Issue #172, opened 2020-11-05
[Neighborhood finder] Looking up a specific address errors out


piratefsh:
## Bug
Looking up `1768 Union Street` on the Neighborhood finder displays an error. The endpoint it hits ` /api/geo/address-metadata`, returns a 500 with an error log of `Address couldn't be found`. 

This address is valid in Brooklyn, so we should look into why this address cannot be found.

Server logs on prod (should be similar to dev):
```
2020-11-05T01:03:10.432892+00:00 app[web.1]: TypeError: Cannot read property 'geometry' of undefined
2020-11-05T01:03:10.432901+00:00 app[web.1]: at exports.addressHandler (/app/src/api/neighborhood-finder/get-address-meta.js:35:9)
2020-11-05T01:03:10.432901+00:00 app[web.1]: at runMicrotasks (<anonymous>)
2020-11-05T01:03:10.432902+00:00 app[web.1]: at processTicksAndRejections (internal/process/task_queues.js:97:5)
2020-11-05T01:03:10.433491+00:00 app[web.1]: POST /api/geo/address-metadata 500 148 - 239.425 ms
2020-11-05T01:03:10.433597+00:00 app[web.1]: Address couldn't be found
```

## Expected behavior
- Endpoint should return a 400 instead of 500 with an error message describing that the address cannot be found.
- UI should display the informative error message

jconard3:
Hi @piratefsh, I took a look at this issue and was not able to replicate the behavior for `1768 Union Street` in neither prod or on my local dev instance. I don't yet have permissions to view prod logs but the post to `/api/geo/address-metadata` returns a `200` in my dev. When you are able, could you please attempt looking up `1768 Union Street` on the Neighborhood finder again to see if the issue persists? Thanks.

piratefsh:
Hi @jconard3, thanks for looking at this! It appears that the issue no longer persists with this address. You can still replicate this when an invalid address is entered, e.g. `sjnjkgjkgs` (keyboard smash a random string).

Right now we return a 500, which just displays an unhelpful message `Error loading. Please try again. If it fails again, let us know in #wg_tech`. We should be returning a 4xx error to indicate that the address is not found (and not a server error) and display a more helpful message. 

We can definitely still use help with issue.

piratefsh:
[...]

PR #184, opened 2021-01-29
Fix for Issue #172 

jconard3:
Return a more helpful message when submitted address cannot be found by googleGeoClient.

Screenshot from 2021-01-28 20-42-26

This fixes Issue #172

### DIFF
--- a/src/api/neighborhood-finder/get-address-meta.js
+++ b/src/api/neighborhood-finder/get-address-meta.js
@@ -27,6 +27,12 @@ exports.addressHandler = async (req, res, next) => {
       },
       timeout: 1000, // milliseconds
     });
+    if (geoResult.data.status == "ZERO_RESULTS") {
+      return res.status(400).send({
+        message: "Zero results returned for submitted address.",
+      });
+    }
+
     const geoResults = geoResult.data.results;
 
     const locResult = geoResults[0];
@@ -67,6 +73,6 @@ exports.addressHandler = async (req, res, next) => {
     );
   } catch (err) {
     console.log(err);
-    return next("Address couldn't be found");
+    return next();
   }
 };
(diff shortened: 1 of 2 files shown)


src/api/neighborhood-finder/get-address-meta.js
```diff
@@ -27,6 +27,12 @@ exports.addressHandler = async (req, res, next) => {
       },
       timeout: 1000, // milliseconds
     });
+    if (geoResult.data.status == "ZERO_RESULTS") {
```
piratefsh, line 30: huh cool! good to know there is a specific status from the API to indicate empty results! nice find.

i think you have a linting error (see below) 👇

src/webapp/pages/NeighborhoodFinder.js
```diff
@@ -17,6 +17,9 @@ import sharedStylesFn from "webapp/style/sharedStyles";
 
 const getError = ({ responseError, responseData, str }) => {
   if (responseError) {
+    if (responseError.response.data.message) {
+      return <>{responseError.response.data.message}</>;
```
piratefsh, line 21: i think you can just return the string here, since its just a string and not a React element.
e.g:
```suggestion
      return responseError.response.data.message;
```

piratefsh (approved):
had a linting error and a small nit, but lgtm and can be merged once those are addressed!

src/webapp/pages/NeighborhoodFinder.js
jconard3: Ah, that might have been an errant `npm run fix`. See https://github.com/crownheightsaid/mutual-aid-app/pull/184/commits/53c0de8f62a7b7bbb269ebdd810a17e586faba88 for a fix.

Issue #181, opened 2021-01-07
[Delivery Map] Remove sms_pickup=true check for displaying sms pickup


piratefsh:
## Summary
We currently hide the sms pickup feature behind a URL param, e.g. `http://localhost:3000/delivery-needed?sms_pickup=true`. This was because we were originally beta testing and didn't want it to appear by default. Now that this is the default way that volunteers are picking up deliveries, we should remove this check and have the `Claim Delivery` button appear by default.

Claim Delivery button should appear by default now:
<img width="308" alt="Screen Shot 2021-01-07 at 12 50 13 PM" src="https://user-images.githubusercontent.com/1868400/103926492-fe0e9200-50e6-11eb-8724-f7ce40f2b0d8.png">

## Code
Where this check happens: https://github.com/crownheightsaid/mutual-aid-app/blob/336aecc4ee18a5c2539c18b4a43cc525a65e9a64/src/webapp/components/ClaimDeliveryButton.js#L12

You may also need to clean up tests if there are any.

PR #185, opened 2021-02-09
Remove sms_pickup=true check 

jconard3:
Remove 'sms_pickup' param used to previously hide a now-implemented beta feature which is now in the default user workflow.

Fixes issue #181


jconard3:
So the button now appears by default upon clicking on a request however I get a slack error message. Is this just a symptom of testing locally?

Screenshot from 2021-02-08 21-51-43

piratefsh (approved):
nice!! thanks for this! yes the Slack error is due to testing on dev as it can't find a Slack instance.

ping me on Slack when this gets merged and I can deploy this for you!

PR #188, opened 2021-02-28
Remove Slack links from "Delivery Needed" map

### LINKED ISSUE #186: Slack links no longer needed in "Delivery Needed" map
Intake volunteers are no longer making Slack posts in the #community_needs channel! The web map is the only method now for listing and picking up deliveries. 

So we no longer need the link back to the Slack post, either in the "Code" column in the web app, or in the pop-up on the map.

Screen Shot 2021-02-23 at 12 25 37 PM
We don't need the link to the Slack post via the neighbor's first name in the pop-up

Screen Shot 2021-02-23 at 12 22 16 PM
We want to get rid of that red error message

Screen Shot 2021-02-23 at 12 22 36 PM
And we don't need the link in the "Code" column there

caroqliu:
Fixes #186

### DIFF
--- a/src/api/delivery-needed/index.js
+++ b/src/api/delivery-needed/index.js
@@ -40,9 +40,6 @@ const {
 const { TWILIO_SMS_DELIVERY_ENDPOINT } = process.env;
 
 const makeFeature = async (r) => {
-  let metaJSON = {};
-  let slackPermalink = {};
-
   const location = await fetchCoordFromCrossStreets(
     `
       ${r.fields[crossStreetFirst]},
@@ -55,27 +52,6 @@ const makeFeature = async (r) => {
     console.error(`[deliveryNeededRequestHandler] could not fetch address location 
       for requestCode: ${r.fields[code]}`);
     return null;
-  }
-
-  try {
-    metaJSON = JSON.parse(r.fields[meta]);
-  } catch {
-    console.error(
-      "[deliveryNeededRequestHandler] could not parse meta",
-      r.fields.meta
-    );
-  }
-
-  try {
-    const channel = metaJSON.slack_channel;
-    const timestamp = metaJSON.slack_ts;
-    slackPermalink = await slackapi.chat.getPermalink({
-      channel,
-      message_ts: timestamp,
-    });
-  } catch {
-    console.error(`[deliveryNeededRequestHandler] could not fetch slack URL
-        for requestCode: ${r.fields[code]} channel: ${metaJSON.slack_channel} and timestamp: ${metaJSON.slack_ts}`);
   }
 
   const need = getDeliveryRequestNeedFormatted(r.fields[supportType], r.fields);
@@ -99,7 +75,6 @@ const makeFeature = async (r) => {
         [timeSensitivity]: r.fields[timeSensitivity],
         [intakeNotes]: r.fields[intakeNotes],
         need,
-        slackPermalink: slackPermalink.permalink || "",
         dateChangedToDeliveryNeeded: r.fields[dateChangedToDeliveryNeeded],
       },
     },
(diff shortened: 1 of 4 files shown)


src/webapp/components/DeliveryTableRow.js
```diff
@@ -83,20 +82,6 @@ const DeliveryTableRow = (props) => {
...
-            <Link
-              href={row.slackPermalink}
-              target="_blank"
-              underline="always"
-              rel="noopener noreferrer"
-            >
-              {row.Code}
-            </Link>
-          ) : (
-            <span>{row.Code}</span>
```
mab253, line 97: hi @caroqliu  - this is awesome! just double-checking here tho - I think we want the code itself to display on the map table, but just not need to be a link. So, it could be just the code text without a link.  Does that make sense?

src/webapp/components/RequestPopup.js
```diff
@@ -67,19 +64,6 @@ const RequestPopup = ({ requests, closePopup }) => {
...
             className={classes.closeIcon}
           />
-          <Typography variant="h6">
-            {meta.slackPermalink ? (
-              <Link
-                href={meta.slackPermalink}
-                underline="always"
-                target="_blank"
-              >
-                {meta["First Name"] || ""}
```
mab253, line 77: same thing here - we want the name to show up, but just not be linked to the slack permalink

src/webapp/components/RequestPopup.js
```diff
@@ -99,21 +83,6 @@ const RequestPopup = ({ requests, closePopup }) => {
             <HouseholdSizeChip size={meta["Household Size"]} />
 
             {meta["For Driving Clusters"] && <DrivingClusterChip />}
-
-            {meta.slackPermalink ? (
-              <DaysOpenChip
```
mab253, line 104: i think the same thing here (tho maybe i'm missing something) - we want the days open/time sensitivity column to still show, just not need to be linked to the slack permalink

piratefsh: +1. for this we just wanna remove the `Cant find slack` error message and always display the `DaysOpenChip`. days open used to rely on the slack post date, but this is no longer true (now it just relies on when the status changed in Airtable)

piratefsh (changes requested):
thanks for opening the PR! however, ditto @mab253 's comments. we want to remove the links to the Slack post, but the keep the text (request code, first name, days open should still show). feel free DM on Slack if we can provide more clarity!

src/webapp/components/RequestPopup.js
caroqliu: Done, thanks for clarifying y'all!

src/webapp/components/DeliveryTableRow.js
caroqliu: Yep, apologies for taking such a haphazard first pass at this 😅  Should be fixed now. Please take another look when you have a moment!

src/webapp/components/RequestPopup.js
caroqliu: Done

src/webapp/components/DeliveryTableRow.js
mab253: looks good to me! (& thank you @caroqliu !!) @piratefsh look good to you?

src/api/delivery-needed/index.js
```diff
@@ -11,7 +11,6 @@ const {
 const { volunteersFields } = require("~airtable/tables/volunteersSchema");
 const { fields } = require("~airtable/tables/requestsSchema");
 const { fetchCoordFromCrossStreets } = require("./fetchCoordFromCrossStreets");
-const slackapi = require("~slack/webApi");
```
piratefsh, line 14: nice cleanup here as well!

piratefsh (approved):
lgtm!! thanks for working on this!

